PR #25, opened 2024-12-26
Add some simple Cucumber Features

rgoulter:
Eventually, I'd like some parts of the smart keymap lib described with "living documentation".

The goal of 'living documentation' is to provide something that's easy to read/understand, while also backed by automated checking/testing. -- This is in contrast to more typical unit/integration tests, which provide automated checking/testing, but are typically difficult to read; or typical documentation, which is not backed by automated checking/testing.

Cucumber one tool to help achieve this.

The executable specifications are written in `.feature` files, with [gherkin syntax](https://cucumber.io/docs/gherkin/reference/).

On the Rust side, each of the steps is bound to some code.

This PR introduces `cucumber` as a dev dependency, and implements some tests which show JSON and RON deserialization to various `key::` `Key` values.

(The large + in the diff count is due to `Cargo.lock` changing).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,76 @@
 version = 4
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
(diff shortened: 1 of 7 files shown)


rgoulter:
These are run by the GitHub action "C Checks" as part of `make test`.

rgoulter:
Rust doc tests may be worth considering; but I think it's much more limited in scope.
https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

May be useful for simple examples, like this deserialization stuff.

PR #34, opened 2024-12-30
Refactor Key::Event associated type to Key<Ev> generic

rgoulter:
This was surprisingly tedious. -- In Rust, the types are expressive enough that they permeate throughout the codebase. So, changing a fundamental type in the codebase will result in many changes.

This PR refactors the `Event` type from being an associated type of the trait `key::Key`, to a generic type of the trait.

Essentially none of the code changed; only the types and generics.

Having `Event` is a step towards allowing multiple implementations of `trait::Key`.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,3 +1,5 @@
+use core::fmt::Debug;
+
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize, Clone, Copy, PartialEq, PartialOrd, Eq, Ord)]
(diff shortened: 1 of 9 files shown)


rgoulter:
Hmm.

The intention for this was to move towards `&dyn key::Key` as the type used in `Keymap`. The hope was there could be lightweight `simple::Key` alongside heavier `layered::LayeredKey(CompositeKey<...>)` or whatever. (i.e. hoping that one key having a heavy size wouldn't affect other keys that didn't need it).

But, with the current `key::Key` needing associated `Context`, etc., and in particular, an associated `PK` for `input::PressedKey<PK, PressedKeyState<PK, Event=Ev>>`... the `PressedKeys` for `&dyn Key` would all need to have the same `K: Key`, which essentially disallows having meaningfully different impl. for `&dyn Key`.

PR #42, opened 2025-01-02
Re-enable the Ceedling Test Suite

rgoulter:
Following up #35.

This PR re-enables the ceedling test suit.

The ceedling tests linked against a custom keymap with 4 keys.

It would be tedious to manually write out `Keys4`, so `seq-macro` is used to generate the `Keys4` tuple struct & `IndexMut` impl. (At the same time, generating all `Keys3` .. `Keys60` dramatically increases compilation time; so, only a select few `KeysN` are generated).

The custom keymap has then been updated to use `Keys4`.

Unfortunately, *1* of the ceedling tests is failing. -- It's not obvious why. There's a Cucumber test which specifies the same scenario, and that passes. (I checked; changing the expected value in the Cucumber feature does fail the test; i.e. the test is checking the expected value). -- I've 'ignored' the test for now, but will dig into why it's failing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -900,6 +900,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "seq-macro"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a3f0bf26fd526d2a95683cd0f87bf103b8539e2ca1ef48ce002d67aad59aa0b4"
+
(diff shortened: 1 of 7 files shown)


rgoulter:
Oh.

The reason the test is failing is a simple one.

#35 made each key *stateful*.

So, this test fails because one of the preceding tests already pushed down the key.

I'll chalk this up to "dealing state is hard".

PR #55, opened 2025-01-04
rust: add some documentation to all items

rgoulter:
For this project, I've done a reasonable job at using tests to drive development.

I've heard the argument that it's good to also write documentation first, since this forces thinking about the design of the interface.

Lots of this effort is wasted:

- There's zero benefit to documenting the `keymap_index` field in a struct named `PressedInput::Key`.
- The Rust docs aren't really the right place to document smart key behaviour. (The interface of the key needs to be documented for users who write keymaps).

However, it's valuable to document the `key` module and the traits in that.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,17 +1,38 @@
 use serde::Deserialize;
 
+/// Input events for [crate::keymap::Keymap].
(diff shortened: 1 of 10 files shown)


rgoulter:
It might make sense to give examples of JSON which the `Key` structs can deserialize from, in each `key::*` mod, and each `key::*::Key` struct.

PR #57, opened 2025-01-05
Refactor layered keys

rgoulter:
Refactoring towards writing Cucumber specs for `key::layered`.

It would be good to eventually construct `key::layered::LayeredKey` using either `[Option<K>; L]` or `Vec<Option<K>>`.

What blocks this is that the `key::LayeredKey` needs to know the number of layers, since its layers are tracked with `[bool; L]` and `[Option<K>; L]`. (And, because Rust is quite strict, this `<const L: usize>` generic percolates up through the types, and so `Keymap` needs to know this, and so the Cucumber `World` struct needs to know..).

I like the idea of: instead of having the `L` be generic, have `key::Context` be generic over the `LayerState` it uses. (For embedded, it can use `[bool; L]` or a `bitset` or whatever. For running Cucumber tests, the `LayerState` can be a `Vec<bool>` (or a `Set<LayerIndex>` or something) which doesn't need to be constrained by some `<L>`.

(As part of this refactoring, I attached some `PhantomData` in `composite::Key`. -- Hence, changed `composite::Key` variants to structs, and added convenience `const` constructors for each variant).

* * *

Though, using `Vec<bool>` isn't possible at this time.

Recall, `Copy` means "can be bit-for-bit copied". And the heap-based `Vec` obviously isn't `Copy`. Hence, the `key::Key` trait should not use `Copy`. -- However, dropping this trait bound is out of scope for this PR.

* * *

Some other thoughts:

Mixed thoughts on Rust.

- In many languages, switching between `[Option<K>; L]`, and `Vec<Option<K>>` would be trivial. With Rust, there's significant friction.

- Though to an extent, I do like using `trait`s to be able to abstract things.

### DIFF
--- a/features/deserialization/key_deserialization.feature
+++ b/features/deserialization/key_deserialization.feature
@@ -23,9 +23,9 @@ Feature: Key Deserialization
   Example: Deserialize a composite::Key (TapHold variant)
     When a composite::Key is deserialized from the RON string
       """
-      TapHold(Key(tap: 0x04, hold: 0xE0))
+      TapHold(key: Key(tap: 0x04, hold: 0xE0))
       """
     Then the result is same value as deserializing the JSON string
       """
-      { "TapHold": { "tap": 4, "hold": 224 } }
+      { "TapHold": { "key": { "tap": 4, "hold": 224 } } }
       """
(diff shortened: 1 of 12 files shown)


rgoulter:
(Also: since `composite::Key` doesn't have any `PhantomData` markers, using struct variants is a bit excessive. -- If more `key::Key`s don't require it, might revert `key::composite::Key` back to using tuple variants).

PR #61, opened 2025-01-06
Chorded Keys

rgoulter:
This PR:

- implements basic chorded key functionality to the Rust keymap.
  - keymap codegen (JSON->RS, and NCL->JSON) to be added later.
- some basic unit tests demonstrating the chorded key behaviour have been added.
- the key has been added to `key::composite`.

It's still rough around the edges; but even `master` branch is a bit rough around the edges atm, and `tap_hold` was merged when it was rough around the edges.

Some thoughts:

- In order to use a nested `key::Key` and its `Context`, practically the easiest thing to do when adding a smart key is to also add its `Event` and `Context` to `key::composite`.

- Although `key::chorded` is added above `key::layered`.. I don't think this means that N smart keys will mean N layers of `key::composite`. -- e.g. "repeat key" clearly wouldn't require another layer; "tap dance" is at the same level as tap hold. "mirrored keys" is at the same level as layer modifier. "leader key" probably requires another "layer" (or re-consideration of how e.g. AuxiliaryKey is impl'd).

- Current implementation is only for 2-key chords. Should be straightforward (albeit tedious) to add 3-key.

- Overlapping/fully-overlapping is still to be implemented. -- Pseudocode for overlapping has been considered.

- The pseudocode I sketched mostly holds up, I think. The pseudocode overlooked the thorniness of the details around resolving the chord.

- Although `key::composite` and its `BaseKey`/`TapHoldKey`/`LayeredKey` are quite unwieldy.. adding `ChordedKey` will helpfully surface common patterns/structure to these. Hopefully that guides smoothing out inconsistencies between them.

- Chorded keys are kinda similar to tap-hold keys, as "key modifiers": when tapped, behaves same as pass-through key. (Similar to how layer behaves when unmodified). Chorded keys modify only when a particular chord of them have been pressed together.

- The no-alloc constraint was tedious to deal with when defining chords in the config. -- I don't want `key::chorded::Config` (nor `Context`) to be generic on `<..., const N>` (since this 'pollutes' its way to `key::composite`; and `key::composite` is used in unit tests for various modules...).
  - `key::layered` also used `[Option<>]`, but there `Option` very nicely suited the intention.
  - Here, instead of `[Option<Chord>]`, I'd rather have `Vec<Chord>`; but `Context` values must be `Copy`.
    - May be, since we're not using `key::dynamic` any more, `&Context` is back on the menu?


rgoulter:
Hmm.

The `key::composite::ChordedKey` enum's `AuxiliaryKey` variant can't be a tuple-struct.

- The derived serde deserialization is for "untagged" enum representation. https://serde.rs/enum-representations.html
  - IIRC, the motivation for using "untagged" is it simplified the NCL keymap-to-json 'serialization' code. IIRC, I changed the repr. to "untagged" when changing `key::composite` to use `BaseKey`/`TapHoldKey`/`LayeredKey` rather than a big `CompositeKey` enum. -- Using "untagged" allowed Cucumber (which is the main consumer of the JSON serialization, other than the JSON-to-Rust codegen) to not have to know about `key::composite`.
  - IIRC, part of the reason I didn't add tags is due to friction for adding it in the keymap-to-json NCL code.

- Previously, using "untagged" enum repr already caused friction; e.g. since `keyboard::Key`'s fields are both optional (a keyboard key can be either a key code, a modifier, a key code plus a modifier..), everything looks like `keyboard::Key`, so the `BaseKey`'s needs to have `Keyboard` as the last variant.

- Here, an `key::composite::ChordedKey`'s `AuxiliaryKey` faces a similar problem: a `key::chorded::AuxiliaryKey` is just a newtype around some other key. So, any key looks like an auxiliary key.

The solution I've gone with is to have the `ChordedKey` variants for `Chorded`/`Auxiliary` be struct variants.

Though, this seems like a dirty kludge than an ideal solution. (Though, I hadn't considered how AuxiliaryKey should be represented in JSON).

It may be that the NCL keymap-to-json supports outputting something that the default externally-tagged serde would support.

PR #82, opened 2025-01-11
Gather custom_keymap types/consts into init module

rgoulter:
It's likely that codegen from `keymap.ncl` will want to generate parts of the initial context. (e.g. for chorded keys, perhaps the chords will be defined in the Context; or perhaps common configuration values can be in the context).

The current `lib.rs` uses two `#[cfg...]` guards. For more items are generated for the `custom_keymap`, more `#[cfg...]` guards would have to be used. -- It's simpler to just have a `mod init { ... }` with `pub` types and consts.


rgoulter:
`cbindgen` fails on this change!?

rgoulter:
cbindgen seems to hate the `pub type Key = ...`; perhaps because many other things in the project share the same identifier (in different modules). -- cbindgen doesn't need to consider those modules, though.

PR #89, opened 2025-01-13
rust: copy Context as values

rgoulter:
I wanted to keep `Context` to be passed as refs.

However, Rust lifetime constraints mean it's difficult to implement changes I want.

Keys can be thought of as either 'base keys' (like `key::simple::Key`), or 'modifier keys' that behave depending on other keys. -- e.g. TapHold is a key modifier; the key either acts as the 'tap' (like a normal key would when tapped), or TapHold allows the key acting as a 'hold'.

Modifier keys must be able to call `new_pressed_key` with the `K::Context` for the inner `K`.

Hence, I want to be able to construct a value like `Pair(Context, K::Context)` from `composite::Context { context1, context2, context3, .. }`. -- With refs, this might be more like `Pair(&Context, &K::Context)`.

In particular, constraints are:

- `key::dynamic::Key` must be an object-safe / `dyn`-compatible trait. In particular, this means the trait's methods aren't generic. (e.g. the `new_pressed_key` can't be generic on some kind of `K::Ctx`).
- Using `&Context` means that `Context` can't contain refs (to anything that lives less than `'static`).
  - Rationale for this constraint is the bounds on `dynamic::DynamicKey`'s `impl` of `dynamic::Key`. We need to be able to make e.g. `simple::Context` (or some other key context) from an aggregate `composite::Context`. Hence, we need something like `&simple::Context: From<&composite::Context>`. -- This bound can't be expressed on the method, because `dynamic::Key` doesn't know about `<K>`. `dynamic::Key` can't know about `<K>`, because the idea is to have an aggregate dynamic type e.g. `DynamicKey<simple::Key>: dynamic::Key<composite::Event>` and `DynamicKey<tap_hold::Key>: dynamic::Key<composite::Event>`. The bound can be expressed on the impl; but, then the LT on the bound is not related to the method. Expressing the LT bound outside the method then requires `for<'c>`, which requires `: + 'static`.
- So `Pair(&Ctx, &K::Ctx)` can't be used as a Context (it's got references).

I considered a few different solutions to try and keep the references in. -- If I can't come up with something, this will have to do.

### DIFF
--- a/src/key/composite.rs
+++ b/src/key/composite.rs
@@ -84,7 +84,7 @@ where
 
     fn new_pressed_key(
         &self,
-        context: &Self::Context,
+        context: Self::Context,
         keymap_index: u16,
     ) -> (
         input::PressedKey<Self, Self::PressedKeyState>,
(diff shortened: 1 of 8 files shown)


rgoulter:
I wondered if passing the e.g. `composite::Context` to the `new_pressed_key` would be a better idea; but, again.. I wanted to try and keep the `key::*::Key` implementations decoupled from one another... -- As far as I could tell, that'd start by having Ctx as a generic to `key::Key`, with bounds `&Ctx: Into<&Self::Context>`. But, I found this impractical by the time I got to `layered::ArrayImpl`, since this requires putting `PartialEq` bound on `Ctx`.

If the firmware runs into efficiency issues, then it'd be a good idea to drop this "implementations decoupled" idea, and just concretely refer to `composite::Context`, I guess.

PR #102, opened 2025-01-17
Impl more NestableKeys

rgoulter:
After updating `key::tap_hold::Key` to `tap_hold::Key<K>`, I realised that for Nickel keymaps to support tap-hold of keys, there need to be more nestable keys.

Implementing Nestable keys for `composite::Event<K>` was high friction. -- I think some refactoring needs to be done before implementing deeply nested keys.

* * *

What I want to see from this PR is unit tests with composite keys with nested keys, for the kinds of keymaps I want.

e.g. For a miryoku-style keymap, the thumb key is layered, but is also used as a tap-hold key with a layer modifier and a simple key.

### DIFF
--- a/ncl/keymap-codegen.ncl
+++ b/ncl/keymap-codegen.ncl
@@ -152,7 +152,7 @@ pub mod init {
     pub type CompositeImpl = crate::key::composite::CompositeImpl<NestedKey, LayersImpl>;
 
     /// Alias for Context type; i.e. [crate::key::context::Context] with generics.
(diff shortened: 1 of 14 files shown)


rgoulter:
It terms of "getting a workable keymap",

https://stevep99.github.io/seniply/
Seniply only needs layers & 'sticky' mods; doesn't need tap-hold keys.

rgoulter:
Hmm, I've worked towards this by refactoring much of the code that touches `Event<E>`. -- I think the changes make good sense, I'll split them into a separate PR.

PR #108, opened 2025-01-18
features.html file, generated from features/ files

rgoulter:
The Cucumber `.feature` files are already pretty close to Markdown. With a little bit of awk (and a few assumptions), these can be translated to markdown, and compiled into an HTML page using pandoc.


rgoulter:
https://rgoulter.com/smart-keymap/features.html

PR #127, opened 2025-01-22
build.rs support SMART_KEYMAP_CUSTOM_KEYMAP=.ncl

rgoulter:
Previously, `SMART_KEYMAP_CUSTOM_KEYMAP` was to point to the generated `keymap.rs` file output from the NCL scripts.

This PR copies/adjusts the "nickel eval" code from the Cucumber test suite, so that the build.rs can read a `.ncl` file from `SMART_KEYMAP_CUSTOM_KEYMAP` and generate the `keymap.rs` for that.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,21 @@
 use std::env;
+use std::fs;
+use std::path::Path;
 
 fn main() {
     println!("cargo:rerun-if-env-changed=SMART_KEYMAP_CUSTOM_KEYMAP");
     println!("cargo::rustc-check-cfg=cfg(custom_keymap)");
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -1,9 +1,91 @@
...
+            Err(io_e) => {
+                panic!("Unhandled IO error: {:?}", io_e)
+            }
+        },
+        Err(e) => Err(e?),
+    }
+}
 
 fn main() {
-    println!("cargo:rerun-if-env-changed=SMART_KEYMAP_CUSTOM_KEYMAP");
```
rgoulter, line 4: Oh, wow; removing this breaks CI. -- Good to know.

PR #143, opened 2025-01-27
TapHold: hold on tap interrupt

rgoulter:
Implements "hold on an interrupting tap".

This only considers a single other key; so, "press TH, press A, press B, release A" still would not resolve as hold.


rgoulter:
This implementation is slightly buggy, and it's fascinating.

PR #165, opened 2025-01-30
Nickel Codegen: construct codegen_values

rgoulter:
This PR iterates towards Nickel codegen support for `LayeredKey<c::TapHoldKey<keyboard::Key>>`. (Currently, instead, it's `LayeredKey<c::TapHoldKey<c::BaseKey>>`).

* Moves from `key_type = fun k => ...`, `rust_expr = fun k => ...` to instead `codegen_values = fun k => { key_type, rust_expr, .. }`.
* Adds several checks in keymap-codegen, covering current behaviour for some tricky cases.

### DIFF
--- a/ncl/keymap-codegen.ncl
+++ b/ncl/keymap-codegen.ncl
@@ -274,6 +274,30 @@ let validators = import "validators.ncl" in
       }
       in
       tap_hold.is_serialized_json serialized_value,
+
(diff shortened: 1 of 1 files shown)


rgoulter:
Looking at CI, seems that `cargo install cbindgen` takes a needless amount of time.

PR #167, opened 2025-01-31
Keymap Pending Key State (Fix TapHold Hold-on-interrupting-Tap)

rgoulter:
Whew, this took a while! -- Hoping to get back to smaller PRs after this.

The problem this PR solves:

- The [ignored](https://github.com/rgoulter/smart-keymap/blob/15da3cd7adb396b3f9106fc5d1395383000c964e/features/keymap/key/tap_hold-config-interrupt-ignore.feature#L53) cucumber feature covers it pretty well.
- With tap-hold hold-on-interrupting-tap, for a sequence like "press (TH(shift)), press A, release A".. the output should be the same as "press shift, press A, release A". -- However, the behaviour didn't reflect this. (The interrupting tap resolves the key to 'hold', but then discards the tap).

The solution to this involves rearranging how `keymap::Keymap` considers pending key state.

This PR has the Keymap have a pending state if a pressed key is 'pending'. (e.g. tap hold key when pressed is 'pending', until resolved as either 'tap' or 'hold'). -- Inputs are passed to the pending key, but otherwise don't affect output until the key resolves.  Once the pending key resolves, the inputs from the pending state are then re-played, affecting output.

(This also adjusts the solution used for tap-hold "tap" output; previously, 'tap' was emulated using virtual pressed keys; that was because the resolved-as-tap tap-hold key was removed as a pressed-input once its key was released. Now, the 'tap' output can be had from the resolved tap-hold key).

(It may be cleaner to refactor the PressedKeyState into "PendingPressedKeyState" and (resolved) "PressedKeyState").

-- Trying this out on a keyboard, it's pretty close. Apparently buggy in some cases.

### DIFF
--- a/features/keymap/key/tap_hold-config-interrupt-ignore.feature
+++ b/features/keymap/key/tap_hold-config-interrupt-ignore.feature
@@ -25,7 +25,6 @@ Feature: TapHold Key (configure interrupt response: ignore)
       }
       """
 
-  @ignore
   Example: rolling key presses (press TH(A), press B, release TH(A))
 
     Rolling the tap-hold key with another key
@@ -50,7 +49,6 @@ Feature: TapHold Key (configure interrupt response: ignore)
       ]
       """
 
-  @ignore
   Example: interrupting tap (press TH(A), press B, release B, release TH(A))
 
     After interrupting the tap-hold key with another key tap (press & release),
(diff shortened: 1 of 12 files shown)


rgoulter:
(When rebasing this after #169, will have to review the diff carefully).

rgoulter:
Trying this out on a KB reveals some cases still don't work well. Will describe these with automated tests.

rgoulter:
Hmm. The behaviour as-is is *closer* to what I want; but:

- The use of `.tick()` & that interplay with the tap-hold resolution is complicated.

- Trying on a keyboard, the code works for simple/happy cases, but still trips up in some cases!

PR #185, opened 2025-02-21
Firmware ch58x clang format

rgoulter:


rgoulter:
Ha.

The `clang-format`, which is *supposed* to be a non-breaking change, broke the CH58x build.

This is because various headers (e.g. `battservice.h`) assume that the HAL's `CONFIG.h` has already been `#include`'d.

And so when `clang-format` arranges the imports, and `battservice.h` is `#include`'d before `CONFIG.h`, types from `CH58xBLE_LIB.h` (such as `bStatus_t`) aren't available.

Ideally, the WCH EVT code would be more robust than this.

PR #187, opened 2025-02-21
Brushup ch32x includes

rgoulter:
I think the `conf.h` was intended to be a single file where various CPP defines are set, before various SDK peripheral includes are handled.

One downside to this approach is that it means *all* the `#includes` are included where the `conf.h` is included. -- Further, the `sdk/`'s `ch32x035.h` then includes `conf.h`, which then includes all the sdk includes. And many of the sdk includes may include `ch32x035.h` itself.

Instead, it's better to manage CPP defines through build configuration (e.g. cmake).

This PR rearranges how things are `#include`d in the firmware/ch32 code. Hopefully each set of includes is more narrow/focused.


firmware/ch32x035-usb-device-compositekm-c/sdk/Debug/debug.c
```diff
@@ -10,8 +10,17 @@
...
+
 #include "debug.h"
 
+#include "ch32x035.h"
+
+#include "ch32x035_gpio.h"
+#include "ch32x035_rcc.h"
+#include "ch32x035_usart.h"
+
+#include "system_ch32x035.h"
```
rgoulter, line 22: This include is in `User/`, because it refers to `SystemCoreClock`, in `DelayInit`.

PR #194, opened 2025-02-25
Add key::chorded to Rust smart keymap library

rgoulter:
Continuing from as #61. -- See discussion there.

- [x] smart_keymap impl.
  - [x] key::chorded
  - [x] key::composite
  - [x] integration tests
- [x] ncl/
  - [x] codegen RS<-JSON
  - [x] JSON<-keymap.ncl
- [x] features/
  - [ ] key
  - [x] keymap

### DIFF
--- a/features/keymap/ncl/chords.feature
+++ b/features/keymap/ncl/chords.feature
@@ -1,0 +1,80 @@
+Feature: Chords
+
+  "Chords" (also sometimes called "combos")
+   allow pressing multiple keys together at the same time
+   to behave as another key.
+
+  For examples of this feature in other smart keyboard firmware, see e.g.:
+
+  - [QMK's combo feature](https://docs.qmk.fm/features/combo),
+
+  - [ZMK's keymap combos](https://zmk.dev/docs/keymaps/combos)
+
+  Background:
+
+    Given a keymap.ncl:
+      """
+      let K = import "keys.ncl" in
+      {
+        chords = [
+            { indices = [0, 1], key = K.C, },
+        ],
+        keys = [
+            K.A, K.B,
+        ],
+      }
+      """
+
+  Example: tapped key behaves as usual
+
+    When the keymap registers the following input
+      """
+      [
+        press K.A,
+      ]
+      """
+    Then the output should be equivalent to output from
+      """
+      [
+        press (K.A),
+      ]
+      """
+
+  Example: tapped together behaves as chord
+
+    When the keymap registers the following input
+      """
+      [
+        press K.A,
+        press K.B,
+      ]
+      """
+    Then the output should be equivalent to output from
+      """
+      [
+        press (K.C),
+      ]
+      """
+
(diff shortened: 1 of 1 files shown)


rgoulter:
Hmm. Might as well add the NCL codegen (JSON->RS, and NCL->JSON) as part of this PR adding `key::chorded`.

rgoulter:
Hmm.

My intuition is that the `unify_cv` logic in the codegen is too simple for what it needs to achieve.

It apparently works in the cases I've tried it for.

Copy-pasting the approach from the `key::layered::Key` codegen for `key::chorded::Key` might work for some cases, but I expect it will have some cases where it does not work.

(Incidentally, I wonder if the codegen'd `keymap.rs` would benefit from including comments with the keymap index, & the JSON value the key was gen'd from).

The `rgoulter` keymap (36-key, 48-key) does have a layered key with both tap-hold base key, and keyboard layered keys.. that's not a trivial case, and the codegen works for that case.

rgoulter:
Just noticed that the cucumber-rs is completely failing, but the GitHub CI check passes (`cargo test` has `0` exit code).

rgoulter:
Ah, okay.

Yeah, the logic related to "unifying" and "lifting" does not support interesting cases.

e.g. chording a layered-key is unsupported. (The Cucumber feature added in the PR is misleading, even if not incorrect).

Roughly, the current code "unifies" the nested keys when constructing codegen values (in tap hold, layered, chorded keys); and when constructing the `key::composite::*Key` (which impl. `key::Key` trait), 'lifts' to suitable types. (The `key::composite::*Key` types typically take in some type which impl `*Nestable`; so, have to be 'lifted').

This approach doesn't work for the case of a `key::chorded::Key` with a `keyboard::Key` chord, and a `key::layered::LayeredKey<keyboard::Key>` passthrough.
(Perhaps it 'could' if all keys were constructed with the fattest data type; but, that increases the firmware size significantly for no benefit). -- In this case, we want the chorded key's nested key type to be `LayeredKey<TapHold<keyboard::Key>>`. The NCL codegen code currently isn't sophisticated enough to handle that.

I think the cleanest thing to do is:
- remove chords from the published docs.
- merge this PR (it 'technically works' for a happy case).
- rework ncl codegen separately, to support this case.

src/key/chorded.rs
```diff
@@ -0,0 +1,766 @@
...
+            keymap_index,
+            key: *self,
+            pressed_key_state: PressedKeyState::new(context, keymap_index),
+        };
+        let timeout_ev = Event::Timeout;
+        let sch_ev = key::ScheduledEvent::after(
+            context.into().config.timeout,
+            key::Event::key_event(keymap_index, timeout_ev),
+        );
+        // TODO: return resolved PKE
```
rgoulter, line 342: Oops, this was left as TODO? Its impl. should be similar to `Key::new_pressed_key`.

PR #197, opened 2025-02-27
RP2040 USB HID TapHold Tap improvements - Mitigate Race Condition

rgoulter:
It seems like there's a race condition in the code. -- I observe a significant percentage of "tap hold misses" (i.e. tap-hold should resolves as tap, but the tap is never reported) with my Pico42 board running the rp2040-rtic firmware. The CH32-48 board does not observe similar misses.

With tap-hold taps, the 'tap' only lasts *1 tick* of the keymap. So, if multiple reports are 'written' before the USB endpoint has sent them to host, the tap would be 'missed'. 

e.g. in the firmware for CH32-48 adapted from the CH32 EVT, the `tick()` is only invoked if the USB transaction has been sent. (A `sending` flag is set when the report is written, which is only cleared when the interface is not-busy; when the interface is not-busy, the sent-report is memcpy'd to the prev-report, which allows the `keymap_tick()` to be called in the TIM3 IRQ handler).

My impression of the Rust code is its behaviour *should* be similar.

This PR mitigates the race condition by introducing a delay between dequeues of queued input events. -- On the one hand, it's really not an elegant solution. On the other hand, it make the library more robust to use with less-than-perfect firmware. -- In practice, it's likely not too impactful on keyboard performance.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -12,6 +12,9 @@ const MAX_SCHEDULED_EVENTS: usize = 32;
 /// Maximum number of pressed keys supported.
 pub const MAX_PRESSED_KEYS: usize = 16;
 
+// Number of ticks before the next input event is processed in tick().
+const INPUT_QUEUE_TICK_DELAY: u8 = 10;
+
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 struct ScheduledEvent<E: Debug> {
     time: u32,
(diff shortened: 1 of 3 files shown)


rgoulter:
Oh. Interesting.

This change surfaces a different bug in the smart-keymap, easily noticeable on the WABBLE-60.

Setting the delay to a higher value makes it easier to reproduce the bug with the CH32 board.

PR #198, opened 2025-02-28
Rust Size Optimization (Nightly)

rgoulter:
Following tips from https://github.com/johnthagen/min-sized-rust

These optimizations bring the firmware size for the 48key-rgoulter keymap with the ch32x firmware down to 52% of flash (compared to 68% before these changes).

AFAICT, removing location-detail is done already (I observed no change to firmware size). Using build-std=core had only marginal impact on firmware size. The 'immediate abort' on panics had a small effect on firmware size.

The lion's share of the size reduction comes from setting `fmt-debug=None`. This requires the nightly toolchain to enact.

It may be that CH32 can't be stuffed with every bell & whistle with a large keymap.


rgoulter:
Not interested in merging these changes at this time.

PR #199, opened 2025-02-28
Enqueue Resolved Key Input at start of Input Queue

rgoulter:
Adding a period between input event dequeues surfaces the awkwardness of the logic.

This PR adds a motivating unit test (passed with INPUT_QUEUE_DELAY set to 0, failed with it set to 10), and adjusts how key resolution enqueues the pending input events.


rgoulter:
And this degrades the Pico42 performance with tap-holds, which the "input tick period" had tried to address.

The keymap code is buggy in a weird way.

PR #208, opened 2025-03-10
NCL: rewrite codegen type unification

rgoulter:
Previously, #194 had carried out the work to add chorded keys to smart keymap. -- However, it was noticed that the chorded keys did not work for *layered keys* in `keymap.ncl`. That practically limits the use of chords, since it can be expected that a smart keymap will have both layers and chords!

The issue was due to `keymap-codegen.ncl`'s poor handling of nested key types.

Going over the context:

- `keymap.ncl` keymap file is first transformed to "serialized json"; and that "serialized json" is then used to generate Rust code when building the smart keymap library.
- The `keymap::Keymap` takes in some `KeysN` tuple of keys. Each key is of a type which implements the `key::Key` trait, with the associated type `PressedKey=key::composite::PressedKey`.
- The `keymap-codegen.ncl` has the Nickel code which manages this Rust code output.

This PR rewrites how it handles nested keys, and unifying the types of those keys. -- The previous implementation *worked*, but was obscure and difficult to work with. (e.g. all `TapHoldKey` types *had to* have `BaseKey` as the nested key; this PR incidentally fixes that!).

Basically, it can get complicated since the nested keys (e.g. the 'tap' and 'hold' keys nested in a tap_hold key, or the 'chord' and 'passthrough' of a chorded key, etc.) must be expressed as the same Rust type, *and* the type must be an item with a suitable implementation of `key::Key`; and since it's an embedded context, we want to keep the values as slim as possible (prefer using newtypes over enums).

The approach taken in the PR is to:
- ensure that the nested keys are constructed as codegen values representing a suitable nestable type. (e.g. the `layered` `codegen_values` ensures its `base`/`layered_keys` are constructed as codegen values representing `TapHoldKey` or the `TapHold` newtype).
- have functions which can transform from the newtype to the enum if necessary.
- keep track of which key::Key implementation is used for each level (base/tap_hold/layered/chorded/etc.)
- be able to lift to the aggregating enum type if necessary.

I don't know if it's harder than it sounds or if it's easier than it sounds; but the code was more complicated to deal with than it needed to be, and now should be much easier to deal with.


rgoulter:
The ch32x build went from 84% to 86%, despite the code being 'simpler'. -- That's annoying.

rgoulter:
In terms of whether this fixes chorded keys with layered keys?

The good news is I can compile & flash a keymap with chords, & invoke a chord.

The bad news is that there are further bugs with the implementation.

rgoulter:
Ah. Seems like #194 did introduce bugs related for the keymap.

rgoulter:
Yeah, it's an issue with memory.

With the `rgoulter` keymap, when I remove the last couple of layers, then I'm able to use chords with the CH32X-48.

rgoulter:
*Might* be worth implementing `key::layered::LayeredKey` with a "sparse keys" implementation.

Rather than having e.g. 6 layers of Option<K>.. a 'sparse' key might cover 6 layers using e.g. 3 Option<K>, and an "index-map".

Or perhaps some other solution.

PR #215, opened 2025-03-12
ncl: keymap-codegen: nestable wraps codegen values

rgoulter:
This PR tidies up some of the mess in the `keymap-codegen` `lift_to_key`.

1. `lift_to_key` was 'smelly' since it had a LARGE `match` expression, which matched over *every* supported key type!

2. the identifier `codegen_values` was used to describe both "'serialized JSON' => codegen values", as well as "codegen values => codegen values" functions.

This PR simplifies the code. Now:
- the `x_nestable` records (`tap_hold_nestable`, `layered_nestable`, etc.) have a function `wrap` which is "codegen values => codegen values",
- each `codegen_values` fun makes use of `key.codgen_values` to construct the nested keys,
- `lift_to_key` is then able to use `chorded_nestable.wrap` and `chorded_.wrap` to construct a value which represents a Rust value which impl. `key::Key`.

### DIFF
--- a/ncl/keymap-codegen.ncl
+++ b/ncl/keymap-codegen.ncl
@@ -224,19 +224,17 @@ let validators = import "validators.ncl" in
 
   # Related to key::composite::TapHoldNestable trait.
   tap_hold_nestable = {
-    # Constructs the codegen values for
+    # Wraps the codegen values for
     #   - key::keyboard::Key
     #   - key::layered::ModifierKey
     #
     # In particular, the returned value is a codegen values record
     #  for an item which implements key::Key with associated
     #  PressedKey = key::composite::BasePressedKey.
-    codegen_values = match {
-      k if keyboard.is_serialized_json k =>
-        keyboard.codegen_values k,
-      k if layer_modifier.is_serialized_json k =>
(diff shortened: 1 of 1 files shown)


ncl/keymap-codegen.ncl
```diff
@@ -866,20 +864,12 @@ let validators = import "validators.ncl" in
...
-          cv |> tap_hold_.codegen_values |> layered_.codegen_values |> chorded_.codegen_values,
-        "crate::key::tap_hold::Key" =>
-          cv |> tap_hold_key.tap_hold_ |> layered_.codegen_values |> chorded_.codegen_values,
-        "crate::key::layered::LayeredKey" =>
-          cv |> layered_key.layered_ |> chorded_.codegen_values,
         "crate::key::chorded::Key" =>
           cv,
         "crate::key::chorded::AuxiliaryKey" =>
           cv,
-        cv =>
```
rgoulter, line 881: Ouch. This badly named identifier was hard to spot!

rgoulter:
This is +1 to "Nickel is neat", "pipe operators are neat", and "this recursive structure is neat".

PR #216, opened 2025-03-13
Rewrite KeyState Pending/Resolved

rgoulter:
## Motivation

Some factors motivating this refactor

- `keymap::Keymap`'s approach to "pending keys" has changed since `key::tap_hold` was written.
  - Initially, `key::tap_hold` took an approach where it would "resolve" on key release; and since the key state didn't stick around once it got released, the tap-hold key then scheduled virtual key presses/releases.
  - Whereas, `keymap::Keymap`'s handling of pending key state is now: it holds the pending key state until events resolve it; then re-plays those events but with the key state resolved.
  - This new keymap pending key behaviour suggests that the PressedKeyState could/should be split into "pending" and "resolved".
- I noticed that the 48key-rgoulter keymap crashes when using two chords, whereas works fine when a couple of superfluous layers are removed. -- This strongly suggests the firmware is running out of memory.
  - `input::PressedKey` has a `key::Key` field. The `key::Key` cannot be replaced with a reference, since its type needs to be known when passed to PKS, and it also needs to be "fat".
    - The size of the "fat" key gets to be quite large! e.g. ChordedKey is going to be 2x the size of fat LayeredKey; LayeredKey is going to be Num-Layers-x the size of TapHoldKey; TapHoldKey is going to be 2x the size of BaseKey. (This gets worse if e.g. adding tap_dance to TapHoldKey, or more key types to BaseKey). -- Exponentially bigger per additional 'key' kind (as such); and number of layers has a linear but significant impact on key size!!

Intuitively, the PK doesn't *need* to store a 'fat' key. -- Each `key::*` can provide a PKS which has all the information it needs. For base keys, that's pretty much the key definition. For tap-hold keys, it's the resolved (nested) key state.

So, rewriting the PressedKeyState ought to help reduce memory usage, as well as providing a cleaner design.

## Summary of Changes

Whew!

This PR rewrites a significant part of the codebase. It rewrites how the `key::Key` trait deals with pressed key state. 

Previously, `key::Key` used an associated `PressedKeyState` type, which followed the tree of `key::Key` impl. types in `key::composite`. (BasePressedKeyState, TapHoldPressedKeyState, etc.). These PressedKeyState would hold typically hold a copy of the key item, as well as tracking whether the key state was 'pending' or not. (e.g. tap hold keys are initially 'pending', before the state resolves to 'tap' or 'hold').

The changes in this PR split this out to an associated PendingKeyState, and a (resolved) key state. (As well, introduces KeyPath so that the nested keys can be referenced).

- This satisfies the PR motivation:
  - I'm able to flash & run the 36key-rgoulter (with all layers), when this seemed to run OOM before this PR.
  - The CH32X-48 build is now 65% (vs 86% on master).

- This dramatically simplifies resolved "pressed key state": pretty much everything resolves to some kind of base key. (key::keyboard, or maybe key::layered::ModifierKey).
- This also clarifies which keys have key state which may be pending. (tap hold, chorded).

One downside to this implementation is it moves a significant chunk of implementation logic to `key::composite`. -- There's a complex / less-than-clear balance of responsibilities between each `key::*::Key` item, and `key::composite`. (The latter is tied to the `key::Key` trait). -- e.g. in particular, the `key::composite::tap_hold` and `key::composite::chorded` `handle_event` are quite heavy.

Some parts of this PR are also a bit looser and more fragile than is ideal. Perhaps some names are less than ideal. (e.g. the case "pending key state resolves to a nested pending key state" is unimplemented; i.e. a chord which is a tap-hold key). I'd rather address that in later, smaller PRs than to keep this big PR open for so long.

## Remarks on Rust

Writing code in memory managed languages like Python is much easier than this.

Rust lifetimes can be difficult. But, rather, the issue is that dealing with stack-based values has limitations; & *especially* doesn't work well with references/pointers.

That's quite a restriction since, in this PR, essentially I didn't want to have (large) `key::Key` impl. struct values tracked as part of key state. Initially, if you don't want to store a copy of `key::Key`, the initial guess is "store a reference to it". -- In unconstrained programming environments, that'd be a good way to address the issue.

I mean. Rust is hard, sure. But, writing code in the no_std embedded context with stack-based values may be hard. -- There's friction which you don't encounter in an unconstrained environment.


rgoulter:
Hmm.

Some of the keymap handling code is a bit awkward with regards to passing events to the pending key state. -- `handle_event` (which is called via `tick()`, and isn't `pub`), has code paths where the pending key state handles the event twice (if the event is an input event).

I don't know if this surfaces as any bugs. -- But, de-cluttering this logic (if necessary) is somewhat out of scope.

rgoulter:
Another awkward case is "pending key state resolves to another pending key state". (e.g. a chorded key which is a tap-hold key). -- I'll defer handling this to another PR.

rgoulter:
I don't think I'll be able to split the diff into smaller commits. -- Well, same as other big PRs; I can commit file by file.

rgoulter:
+1,018 −1,206 even without codegen snapshot. Oof.

rgoulter:
CH32X CI job, flash down to 62.34% (vs 86% on master).

rgoulter:
The "Pressed Key Events -> ScheduledEvent -> Event::Key" is higher friction than it needs to be, for the (common) case of "I have some key::*::Event, and want to construct PKE with that".

rgoulter:
Tried this on the CH32X-48 board:

1. Need to add a test for chording for either key press order. (e.g. for chord AB=>, test "press A, press B" and "press B, press A".

2. Certain tap-hold keys (not backspace?) seemed to be broken. (Didn't report output, couldn't report output with other keys). Though, chorded key passthrough worked.

rgoulter:
The failing test makes sense. -- Sophisticated nesting of keys requires care.

In the motivating test case, it's a layered key of a tap-hold key.

The code assumes that layered key does not have any pending key state, since it directly resolves to its nested key state.

However, in order to pass events to the `key::Key::handle_event` for the pending key state, the `&Key` used *is* the layered key!

(Similarly, although 'chorded key passthrough tap' is covered by test, 'chorded key of a tap-hold key' would fail).

The right 'fix' here is to have a PendingKeyState for LK which manages *which* key the LK resolved as. (Base or layer number). -- When the nested PKS resolves, then the layered::PKS resolves.

src/key/mod.rs
```diff
@@ -392,84 +427,27 @@ impl KeyOutput {
...
-        &mut self,
-        context: Self::Context,
-        event: Event<Self::Event>,
-    ) -> PressedKeyEvents<Self::Event>;
-
-    /// Output for the pressed key.
-    fn key_output(&self) -> KeyOutputState;
-}
-
+// TODO useful?
```
rgoulter, line 430: Apparently not.

rgoulter:
Current status:

- The bulk of this PR, which replaces the tree of "xPendingKeyState" nodes with `key::composite::KeyState` and `PendingKeyState` seems pretty good.

- The first regression (mentioned above) is that tap-hold keys broke within layered keys. That was because the tap-hold key's pending state tried to use the layered key to update, but needed to use the tap-hold key.

  - The solution which seemed most promising was to refer to `&dyn key::Key` in the `PressedKeyResult::Pending` variant.
  - However, this *quickly* runs into issues with lifetimes in `Keymap`: keymap's pending key state borrows this `&dyn Key`; but the ref refers to a key in `&self: Keymap`. But, since `&self` is borrowed in the pending state, `&mut self` cannot be used.
    - The current solution on the branch works around that by having `Keymap` take in `&keys` rather than `keys`.
- The current issues this branch face:
  - the change in `keymap` needs to be applied to codegen. Trivial.
  - Cucumber's `World` doesn't interact well with items with lifetimes.
  - The chord unit test still fails.

rgoulter:
https://stackoverflow.com/questions/32300132/why-cant-i-store-a-value-and-a-reference-to-that-value-in-the-same-struct

rgoulter:
I wonder if one way to avoid `& dyn key::Key` in `key::PressedKeyResult::Pending` is to instead store a path or index. e.g. rather than `usize` (which gets the whole key definition), also pass across the resolution (`.passthrough .passthrough .tap`).

If dealing with the lifetimes/references in Cucumber causes too much trouble, I might have to opt for that.

rgoulter:
Implemented lookup key using a "KeyPath".

The implementation of it is quite 'fragile'.

e.g. in the same way that the code for the chorded event uses `bool` to indicated "chord resolved to chord or to passthrough" is fragile.. here, the KeyPath is a Vec of u16. -- Ideally, this can be refactored later.. but, using `u16` itself allows it to be used in `key::Key` without fussing over *another* associated type.

One way the impl. as-is is fragile is e.g. indicating ".tap" with `0`, ".hold" with `1`. -- But the bigger fragility is the assumptions about the state of the `KeyPath=Vec<u16, ...>`. Currently, the code assumes that each time `nested.new_pressed_key` is called from a key with non-trivial pending key state, that the keypath is updated. -- This is the opposite of "make bad states unrepresentable".

The firmware works; but the tests are still to be updated.

Anyway, this has eliminated the need for storing `&'k dyn key::Key` in Keymap, which ultimately blocked testing the code in Cucumber.

src/key/composite/chorded.rs
```diff
@@ -51,104 +45,251 @@ pub struct Chorded<K: ChordedNestable>(pub K);
...
+    }
+
+    fn handle_event(
+        &self,
+        pending_state: &mut Self::PendingKeyState,
+        context: Self::Context,
+        key_path: key::KeyPath,
+        event: key::Event<Self::Event>,
+    ) -> (Option<Self::KeyState>, key::PressedKeyEvents<Self::Event>) {
+        let keymap_index: u16 = key_path[0]; //
```
rgoulter, line 66: Dangling `//`.

src/key/composite/tap_hold.rs
```diff
@@ -31,89 +36,170 @@ pub enum TapHoldKey<K: TapHoldNestable> {
...
     }
+
+    fn handle_event(
+        &self,
+        pending_state: &mut Self::PendingKeyState,
+        context: Self::Context,
+        key_path: key::KeyPath,
+        event: key::Event<Self::Event>,
+    ) -> (Option<Self::KeyState>, key::PressedKeyEvents<Self::Event>) {
+        let keymap_index = key_path[0]; //
```
rgoulter, line 67: Ditto.

rgoulter:
I've rewritten the commits. Having 1 squashed commit with +/- 1300 is has advantages, but makes it somewhat difficult to read.

PR #219, opened 2025-03-19
rust: keymap: use Index, not IndexMut

rgoulter:
An improvement surfaced while working on #216.

Since the dynamic key is no longer the mutable "dynamic key" impl., Keymap only needs to use `Index` not `IndexMut`.


rgoulter:
Hmm. If ever want to support dynamic keymap, then *may* want indexmut.

With immutable keys (which is *good* to have), dynamically adjusting keymap would involve mapping/replacing the `keymap` instance.

PR #227, opened 2025-03-22
keymaps: add 75-key keymap

rgoulter:


tests/ncl/keymap-75key-rgoulter/keymap.ncl
```diff
@@ -0,0 +1,69 @@
+let K = import "keys.ncl" in
+
+let RAISE_  = 1 in
```
rgoulter, line 3: Right.. this is a confusing aspect of the smart-keymap implementation:

layer index 0 refers to the first layer *above* the base layer.

PR #229, opened 2025-03-23
Split Keyboard support

rgoulter:
Although "split keyboard" is really a "hardware" feature rather than a keymap feature, I think it'd be useful to provide functionality to support the split keyboard implementation.


rgoulter:
Hmm. Trying this on CH32X-36. It doesn't reliably work.

rgoulter:
I've tested this with code for the Minif4-36, with full-duplex UART. (#238).

PR #231, opened 2025-03-23
Remove "base" from LayeredKey

rgoulter:
Although I like the idea of "base key & [additional layers]".. -- the layer indices then need to be offset by 1.

This PR is an attempt to see if removing base key would simplify that. (Otherwise, 'base' is always enabled; and arrays can otherwise be 1-based indices).


rgoulter:
#232 addresses confusion about layer index in `keymap.ncl`.

And I do like `LK = { base, layered }` conceptually.

PR #232, opened 2025-03-24
keymap.ncl use more intuitive layer index

rgoulter:
Because `key::layered::LayeredKey` is implemented with `{ base, layered }` (i.e. some non-optional base key, plus a sequence of optional layered keys), layered key index `0` refers to `layered[0]`, i.e. the first non-base key.

There's inevitably some kind of "off by 1" going on here.

I think a decent solution to this is to have layer indices in `keymap.ncl`'s `layers` use "layer index `0` refers to base", and transform/shift the layered modifier key layer index by 1 when building the keymap code.


rgoulter:
One point of friction in this PR is that the `the keymap registers the following input` cucumber step... it's a bit

rgoulter:
Hmm.

While there does have to be "off by one" somewhere... The implementation as-is (modify layer_modifier's layer_index in the keymap-ncl-to-json) is apparently a really nasty place to put it.

I think instead it'd be "cleaner" to change the layer modifier argument to be a "layer index" wherein "1 = first non-base layer", and update the Rust impl. to use 1-based indices.

rgoulter:
Yeah, okay. This is a much nicer impl.

By having "layer index 0 means base layer" consistently everywhere, this makes it more intuitive to write `keymap.ncl`. Using `LK = { base, layered }` still means dealing with "off by 1" somewhere; but this narrows it down to dealing with the index in the implementing structs.

PR #233, opened 2025-03-24
Layer Modifier Key: Set Default Layer

rgoulter:


rgoulter:
The code works.

But the code also lacks a Cucumber feature spec (for either the 'set default' key itself, or for use in a `keymap.ncl`); the supporting NCL code has no checks, nor have any unit tests been added. -- Add these before merging.

rgoulter:
Also: found a bug with the codegen: if the layer modifier keys are on a layer alongside the tap-hold keys, the code doesn't compile.

PR #237, opened 2025-03-26
STM32f4 rtic:  minif4 36 rev2024 1

rgoulter:
This PR copies across code for the MiniF4-36, rev 2024.1. (Because I've got a build of MiniF4 rev2024.1 which supports full-duplex UART).

Changes to the code from keyboard-labs:
- the same `mod board` abstraction, that the `main.rs` uses in the `-rtic-smart-keyboard` crates.
- changes the split transport code to encode `smart_keymap` `Event`, rather than `Keyberon` events.
- minor updates for newer STM32F4xx HAL.

Stuff like that the STM32F4-RTIC code assumes UART1 is used as the split transport does limit usefulness; but it's not a priority to improve limitations like that.

### DIFF
--- a/stm32f4-rtic-smart-keyboard/src/bin/minif4_36-rev2021_4-lhs.rs
+++ b/stm32f4-rtic-smart-keyboard/src/bin/minif4_36-rev2021_4-lhs.rs
@@ -1,0 +1,260 @@
+#![no_main]
+#![no_std]
+
+#[cfg(not(custom_board))]
+mod board {
+    pub use stm32f4_rtic_smart_keyboard::input::Input;
+
+    pub use stm32f4_rtic_smart_keyboard::app_prelude::VID;
+
+    pub const COLS: usize = 5;
+    pub const ROWS: usize = 4;
+
+    #[rustfmt::skip]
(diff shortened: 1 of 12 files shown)


stm32f4-rtic-smart-keyboard/src/bin/minif4_36-rev2021_4-lhs.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub const PID: u16 = 0x0005;
+    pub const MANUFACTURER: &str = "smart-keyboard";
+    pub const PRODUCT: &str = "MiniF4-36 rev2021.4 LHS";
+
+    pub type Keyboard = usbd_smart_keyboard::input::Keyboard<COLS, ROWS, Matrix>;
+
+    pub type PressedKeys = usbd_smart_keyboard::input::PressedKeys<COLS, ROWS>;
+
+    pub struct Matrix(
+        pub keyberon::matrix::DirectPinMatrix<hal::gpio::ErasedPin<hal::gpio::Input>, COLS, ROWS>,
```
rgoulter, line 30: `hal::gpio::ErasedPin<hal::gpio::Input>` could probably be named "InputPin". IIRC, the rp2040_rtic code has a similar type alias.

PR #239, opened 2025-03-27
CH32X firmware: add keyboard ch32x-36 (split keyboard)

rgoulter:
Carrying on from #229.

* * *

This PR adds support for split keyboards for the CH32X firmware. More specifically, an implementation using full-duplex USART2, with DMA for RX & TX.

It had taken some time to troubleshoot issues I was having with the CH32X-36. -- Somehow, in some cases, it's sending the same input event twice. -- The keymap logic was updated to mitigate this case. I've not observed any issues since then.

The implementation is only for full-duplex USART2. I expect it to be straightforward to add support for other USART, perhaps even for alternate function / remapped USART.


rgoulter:
https://github.com/rgoulter/keyboard-labs/pull/20

rgoulter:
Getting TX4 working on PA5 was very helpful for troubleshooting issues.

My initial implementation was to write 1 byte at a time from the 4 bytes of the serialised `Message`, received using the IRQ for RXNE.

This basically didn't work for some reason.

Troubleshooting this, the issue appeared to be that of the 4 bytes written, even though each byte waited for the USART TC register to set, only 2 bytes were received. (Of 1,2,3,4, the receiver received 1,2). -- I observed this by having code send 4 bytes once every second, and observing which were received.

I tried mitigating this by using DMA with USART. I observed no difference when using DMA with TX2, but the issue seemed mostly improved when using DMA with RX2.

Using RX DMA with buffer length 4 and IRQ DMA1_Channel6_IRQn *mostly* works; but runs into issues when keys are pressed too closely together. :| (e.g. for a chord). -- I think will have to IRQ on USART_IT_IDLE (like the evt example 'dma usart idle recv'). -- This would allow sending/receiving messages that aren't just 4 bytes long, which would be good to have anyway.

* * *

Besides this: smart keymap lib. does not gracefully handle the "keep receiving input when its pressed_keys is full".

rgoulter:
The UART communication now uses DMA for receiving bytes; IRQs on DMA TC and USART IDLE will swap the DMA buffer, & receive the bytes (deserialise into messages).

Using DMA with a buffer larger than 4 bytes for RX significantly improved reliability, especially compared to just reading 1 byte at a time.

-- Just observed a bug, somehow, where the keyboard half receives the correct press/release event, but the keymap isn't pressing the key. Huh.

Anyway.

Next steps:

- Supporting "split keyboard or not" requires a different approach for the NCL / codegen / CMake / C code. e.g. CH32X wants to use TX4 for debug. -- Generating *all* the C code from Nickel expressions seems silly. -- Using Nickel to generate config options and *maybe* inlining stuff like matrix scans seems okay.

- This demo/poc code needs to be brushed up a bit; no need to have the values set on split = TX4.

rgoulter:
After using it a bit more, the firmware still has some rough edges. -- Will require further investigation.

rgoulter:
Ah, okay.

Somehow, the keyboard LHS is receiving the same event twice.

`keymap::Keymap` does not gracefully handle this. It *should* ignore the a press event for an already pressed input. -- As is, it gets into a bad state, where multiple values with the same keymap_index are pressed. -- This is mitigated by checking for the keymap index when processing the press event.

Still, I want to try to see if I can find *why* it's receiving the same event twice.

PR #240, opened 2025-03-29
Feature: layer "default" layer modifier key

rgoulter:
This PR adds & implements a Cucumber spec for the "set default layer" modifier key.

The somewhat-cumbersome `ncl/layered-key.ncl` code is updated to work with "layer state", rather than just "active layers".

### DIFF
--- a/features/keymap/ncl/layers-default.feature
+++ b/features/keymap/ncl/layers-default.feature
@@ -1,0 +1,39 @@
+Feature: Layers (Default Layer)
+
+  The `K.layer_mod.set_default` key allows setting the Default layer.
+
+  Background:
+
+    Given a keymap.ncl:
+      """
+      let K = import "keys.ncl" in
+      {
+        layers = [
+          [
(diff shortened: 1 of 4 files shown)


features/keymap/ncl/layers-default.feature
```diff
@@ -0,0 +1,39 @@
...
+      """
+
+  Example: tapping the set default layer modifier key changes the default layer
+
+    When the keymap registers the following input
+      """
+      [
+        press (K.layer_mod.set_default 1),
+        release (K.layer_mod.set_default 1),
+        press (K.B),
```
rgoulter, line 33: 1. "tap" rather than "press", "release" may be nicer.

2. "press K.B" is rather hard to read in this case.

rgoulter:
I'll have to think about the `press`/`release` syntax.

May be want something more like `press key(K.B)` / `press key_on_active_layer(K.B)`. May want `press(key_at_index(2))`.

PR #241, opened 2025-03-29
Fix keymap.ncl codegen bug

rgoulter:
It's subtle; but that the `key_type` record in the codegen values used the `cv` argument rather than the `nested` field resulted in a bug. -- The `nested` field may be updated.

Motivating test case added (& fixed).


rgoulter:
I'm curious if an LLM assistant could have found the bug. I think if you narrowed it down to the motivating test case, showed the compilation error, and showed the code, then there's a good chance it could have. -- The discrepancy between use of `cv` and `nested` in the `composite.tap_hold_key.tap_hold_` fun. is probably something it would notice.

PR #242, opened 2025-03-29
bugfix: handle double key release

rgoulter:
I'm observing cases where a key press received from a split-RHS, but doesn't result in a key press.

My first guess is that this is because a "key press" is being missed, and the keymap doesn't handle (extra) key releases well.


rgoulter:
Hmm. That test case doesn't reproduce the issue.

PR #244, opened 2025-03-29
Fix keymap tick-less input handling

rgoulter:


rgoulter:
FWIW, Not all overflows are ignored... e.g. if more than max_pressed_keys are pressed at once, the firmware will still panic. I'm not sure how 'robust' I want the keymap firmware to be against all these cases.

PR #249, opened 2025-03-30
Rust, transform values, remove unused methods

rgoulter:


rgoulter:
Not sure what I expected from GH Copilot's PR reviews. Interesting!

PR #251, opened 2025-03-30
Key State Persistence

rgoulter:
This PR:

- Adds `key::KeyState::is_persistent`.
- Implements Keymap functionality in `process_input` to handle persistent key state.

"Key State Persistence" means the key state stays pressed even when the key is released.

I expect this will be useful for things like caps word, "sticky" modifier keys, and tap-dance keys.

There's some overlap with "pending" key state here.. -- Depends on perspective; but when you press a key with pending state (tap-hold, chorded, sequence), the current implementation is that it's state is "pending" until it resolves to a key state. -- An alternative implementation might have it immediately be a pressed key state, where the pressed key state itself handles whether it's resolved or not. -- Perhaps if I implemented sticky keys before tap-hold keys, that's the approach I would have taken!

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -24,6 +24,9 @@ pub struct PressedKey<S> {
     pub keymap_index: u16,
     /// The pressed key state.
     pub key_state: S,
+    /// Whether the key is currently pressed.
+    /// (Persistent key state can be retained even when key is not pressed).
+    pub key_pressed: bool,
 }
 
 impl<Ctx, Ev, S: crate::key::KeyState<Context = Ctx, Event = Ev>> PressedKey<S> {
--- a/src/key/composite.rs
+++ b/src/key/composite.rs
@@ -262,6 +262,10 @@ impl key::KeyState for KeyState {
             KeyState::NoOp => None,
         }
     }
+
+    fn is_persistent(&self) -> bool {
+        false
+    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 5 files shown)


rgoulter:
Oh, wow.

Copilot's low-confidence suggestion "There appear to be two separate loops processing events on pressed_inputs (one using ev.into() and another directly with ev). Verify if both loops are necessary or if this might lead to duplicate event handling." is a pretty good catch, I think.

(I'd noticed this before; but it's not been a problem so far. Might be a problem with tap-dance).

rgoulter:
> Depends on perspective; but when you press a key with pending state (tap-hold, chorded, sequence), the current implementation is that it's state is "pending" until it resolves to a key state. -- An alternative implementation might have it immediately be a pressed key state, where the pressed key state itself handles whether it's resolved or not. -- Perhaps if I implemented sticky keys before tap-hold keys, that's the approach I would have taken!

On the other hand.. I think the case "resolve as hold on interrupting tap" requires delaying (or replaying) input events in the keymap, anyway.

rgoulter:
Hmm.

I'm not so sure about this.

e.g. a motivating use cases are: the Caps Word key, Sticky modifier keys. -- e.g. Tapping the caps word key persists a caps word state, which acts as shift, until the caps word key is deactivated.

One concern I have is poor interaction with complex keys which have pending key state. -- e.g. tap-hold keys, or chorded keys.

e.g. for caps word toggle, one use case is hitting the 'toggle' key to disable caps word. -- But, if caps word is implemented as a persistent key state (the resolved key state, which persists after the key is released), then some assumptions are in conflict. -- Key state wouldn't know that it's handling an event for *its* key, even if the keymap index is the same. In order to figure out the resolved key state is the same, a bunch of complexity would have to be added. Smells off.

PR #253, opened 2025-04-01
Add CapsWord Key

rgoulter:
"Caps Word" is similar to "Caps Lock", but auto-deactivates when a non-word key is pressed. This makes it useful for typing out constants in programming languages.

This PR implements Caps Word by using a persistent key state which outputs the shift key (and deactivates when e.g. Space is pressed).

Builds upon persistent key state from #251.

There may be other ways to implement Caps Word. e.g. Maybe the keymap supports some kind of "output filtering"/"output mapping". I haven't thought about how smart-keymap might implement the fancy "magic key" or "adaptive key". -- But, in any case, it's better to have *an* implementation now, than a better implementation later.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -24,6 +24,9 @@ pub struct PressedKey<S> {
     pub keymap_index: u16,
     /// The pressed key state.
     pub key_state: S,
+    /// Whether the key is currently pressed.
+    /// (Persistent key state can be retained even when key is not pressed).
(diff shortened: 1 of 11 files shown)


rgoulter:
Need to re-implement this some other way.

rgoulter:
Can't re-open a PR after it's been force-pushed.

PR #258, opened 2025-04-02
Context emit events

rgoulter:


rgoulter:
`PressedKeyEvents` somewhat of a misnomer. `KeyEvents` or `Events` would be clearer.

PR #259, opened 2025-04-02
Add Key: Caps Word

rgoulter:


rgoulter:
Tried on the keyboard. Works.

PR #261, opened 2025-04-03
Sticky Modifier Keys

rgoulter:


rgoulter:
`keymap::Keymap` doesn't regard modifier virtual keys as any different than regular virtual keys.

But, sticky keys will combine e.g. "Ctrl" and "Shift" into a VK "Ctrl+Shift".

Most likely, releasing VK modifier should affect all VKs.

Not sure whether it's better to impl. that in this PR, or separately.

rgoulter:
The report equivalences: `[[2, 0, 4, ...], ...]` is being treated differently than `[[2, 0, 0, ...], [2, 0, 4, ...], ...]`, even though these are practically the same.

So long as that's the case, can't add Cucumber spec for many sticky modifier key cases.

rgoulter:
Firmware size: 72.90% -> 77.39%

rgoulter:
I'd worried that increasing PKE size would have a dramatic effect. It doesn't.

rgoulter:
I *think* it's reasonable to limit the number of sticky modifier keys to "4 at once". That's enough for one of each LHS modifier.

For use cases which require sticky modification of more than that, it'd be possible to cover all 8 modifiers with up-to 4 keys.

Worst case, I can move this `const MAX = 4` to `crate::init` & include it in codegen.

rgoulter:
CH32X now at 77.74%.

src/key/sticky.rs
```diff
@@ -0,0 +1,293 @@
...
+
+/// Sticky Modifiers context.
+#[derive(Debug, Clone, Copy)]
+pub struct Context {
+    /// The sticky modifier key configuration.
+    pub config: Config,
+    /// Sticky modifiers.
+    pub active_modifiers: [key::KeyboardModifiers; MAX_STICKY_MODIFIERS as usize],
+    /// Number af active sticky modifiers.
+    pub active_modifier_count: u8,
```
rgoulter, line 51: `pub` here is gross. -- Is this necessary? `from_config` should be enough.

rgoulter:
Heh. Trying this out, the default settings don't work nicely with rolling; e.g. "sticky shift, press a, press b, release a".. this shifts both A and B rather than just A. -- Still, it's a good start.

rgoulter:
Symbols aren't being modified (shifted) with this. (e.g. `/`, `;`)

PR #263, opened 2025-04-04
Virtual Keys use key::KeyOutput

rgoulter:
Virtual Keys were initially implemented with just `u8` as the virtual key code.

This PR updates them to take in a `key::KeyOutput` value.

The motivation for this is so that Sticky keys will be able to use virtual keys, with more than one modifier key. (This is simpler than having sticky keys manage/emit multiple virtual key events).

### DIFF
--- a/smart_keymap/src/lib.rs
+++ b/smart_keymap/src/lib.rs
@@ -35,7 +35,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use smart_keymap::{init, input, keymap, split};
+use smart_keymap::{init, input, key, keymap, split};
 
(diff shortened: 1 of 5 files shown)


smart_keymap/src/lib.rs
```diff
@@ -59,6 +59,8 @@ pub enum KeymapInputEventType {
 }
 
 /// Input event.
+///
+/// LIMITATION: split transport for virtual keys only works for basic key codes.
```
rgoulter, line 63: The intended use for the `KeymapInputEvent` in `libsmart_keymap` is for split matrix events.

Virtual Keys are more of an implementation detail for some of the smart keys.

PR #267, opened 2025-04-05
ci: add nickel-format CI job

rgoulter:


rgoulter:
Tested with a commit which requires formatting. The CI job indicates the code needs formatting, as expected.

PR #275, opened 2025-04-05
Refactor: decouple key::composite from key::chorded

rgoulter:


rgoulter:
0/2 for copilot, here. -- The first answer is a *good* question/concern to point out; the second is not.

PR #276, opened 2025-04-06
Decouple key::Key impls from key::composite

rgoulter:
Recall, the previous big refactoring of `key::Key` impl. involved removing the dyn compatible `key::dynamic::Key`, making `key::Key` dyn compatible, moving `key::Key` impls. out from each `key::*` module, and putting more stuff in `key::composite`.

I liked the benefit of each `key::*` being simpler. But, I don't like how `key::Key` is *only* implemented in terms of `key::composite`. -- Want to allow (in theory) that some other aggregate key types could be used. (Well, once `keymap` is genericised).

This PR addresses that.

* * *

I spent some time considering ways to have `key::Key` impls not depend on `key::composite`.

One point of friction is that the `key::Key` implementers, such as `key::keyboard::Key` don't specifically refer to Context, Event, PKS, KS (the associated types of `key::Key`). -- So, in order to have a `key::Key` impl for `key::keyboard::Key` generic over these, `PhantomData` would have to be introduced somewhere.

One option was having phantomdata in a newtype alongside `Key`. -- Rename current `Key` items to `KeyDefinition`, and add a newtype-ish struct with that key definition & phantomdata. -- I rejected this option since it would involve too much fuss over the generics for these new `Key` types. (The generics would have to be added to codegen, which I felt would make it too verbose).

Another option might have been to use generics for the `key::Key` trait rather than associated types. This would allow `key::Key` impls without e.g. `key::keyboard::Key` needing phantomdata, but, this just shifts the requirement for phantomdata around (e.g. `Keys1` would require phantomdata).

Ultimately, I decided on a compromise: use the `Context`, `Event`, etc. from `crate::init`. -- Since `crate::init` is codegen'd, this would (in theory) allow changing the associated types for the `key::Key` from `key::composite` to some other implementation.

The `key::composite` code is now boilerplate, as it should be.


rgoulter:
Some things I'd assumed would be zero-cost, aren't. -- It'd be nice if "impl. Deref" was zero-cost, but it's not. -- It's useful to have for tests, though.

rgoulter:
This PR increased the CH32X-48 firmware size a bit, too.. despite being a plain refactoring.

rgoulter:
Hmm.

But `key::composite::TapHoldKey` *already* refers to the sum type of "tap hold | BaseKey".

So, using aliases is "cute", but I'm not sure if it can get around that collision.

rgoulter:
If not using type aliases, what about using `KeyImpl` instead of `Key`?

rgoulter:
Oh.

Recall some

rgoulter:
Oh.

Recall, in Rust, trait generics vs associated types... consider these two examples: this is valid:

```rust
trait Foo<T: core::fmt::Debug> {
    fn print_foo(&self, t: &T);
}

struct S;

impl<T: core::fmt::Debug> Foo<T> for S {
    // ...
}
```

whereas, this is not:

```rust
trait Foo {
    type T: Debug;
    fn print_foo(&self, t: &Self::T);
}

struct S;

impl<T: Debug> Foo for S {
    type T = T;
    // ....
}
```

As it fails for the familiar error:

```
the type parameter `T` is not constrained by the impl trait, self type, or predicates
```

In my codebase, `key::Key` has associated types for Context, Event, PressedKeyState, KeyState.

Currently, `key::Key` is implemented only for `key::composite::Context`, etc. 

In order to have a generic impl. of `key::Key` for `key::keyboard::Key`, some kind of `KeyImpl` with PhantomData would be required, since `key::keyboard::Key` makes no direct mention of Context, Event, etc.

Associated types felt like the more natural option over generics. But, having Context, Event, etc. as generics to the trait would provide a way of linking these generics to the implementation. -- This seems like a very subtle point to me.

PR #278, opened 2025-04-07
Rust: add bin/sizeof

rgoulter:


rgoulter:
https://github.com/rgoulter/smart-keymap/actions/runs/14299695606/job/40071936306?pr=278 running into "no such file or directory"?

rgoulter:
Probably this (& the CH32X firmware build) which output the size would benefit from some kind of CI step which compares the result against previously known size, and remarks on increase/decrease.

PR #282, opened 2025-04-08
keymap: emit ResolvedKeyOutput for resolved pending state

rgoulter:
Was impl'd as part of #261.


rgoulter:
Ideally, cover this with at least an integration test of sticky keys with chorded/tap-hold keys. (Using a sticky key to modify a tap-hold key requires this PR).

PR #296, opened 2025-04-12
ncl: add contact for config for keymap.ncl

rgoulter:


rgoulter:
Hmm.

Nickel keeps the contract on `config.chorded`, even when we try to merge/replace it with other things. That's annoying. -- Not sure how to work around it.

ncl/keymap-ncl-to-json.ncl
```diff
@@ -518,14 +555,20 @@ let validators = import "validators.ncl" in
...
+      let { chorded = km_config_chorded, ..km_config } = config & { chorded = {} } in
+      let chord_indices = chords |> std.array.map (fun { key, indices, .. } => indices) in
+      let config_json_value =
+        km_config
         & {
-          chorded.chords = chords |> std.array.map (fun { key, indices, .. } => indices),
+          chorded =
+            km_config_chorded
+            & {
+              chords = chord_indices,
```
rgoulter, line 566: Hmm, -1 to Nickel here!

Nickel treats `config` and `let { chorded = km_chorded_config, ..km_config } in km_config & { chorded = km_config_chorded }` differently with regards to what contracts are applied on them.

Or, even these have different contracts, apparently: `config.chorded`, `let { chorded = km_chorded_config, .. } = config in km_chorded_config`.

Otherwise, this diff should be 'pure'; but, this diff fixes an error I was getting.

PR #299, opened 2025-04-12
DistinctReports modifier press/release equivalences

rgoulter:
The Cucumber spec scenarios "tapped sticky modifier keys stack", "sticky modifier key acts as regular key when interrupted by tap" require the DistinctReports to equate different sequences of reports depending on how modifiers are reported.


rgoulter:
This also only works for N=1 modifier.

A better approach may be to filter out these reports in the `Eq` implementation.

PR #302, opened 2025-04-13
Add stm32-embassy-smart-keyboard firmware example

rgoulter:
I want to provide an example firmware for split keyboards which have single-wire serial between split halves.

I have builds of the MiniF4-36 which have TX-TX; this can use half-duplex UART.

`stm32f4xx-hal` doesn't support half-duplex uart at the time of writing. `embassy-stm32` does. So, I'll add an embassy example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2213,6 +2213,7 @@ dependencies = [
  "embassy-usb",
  "keyberon",
  "keyberon-smart-keyboard",
+ "panic-halt",
  "panic-probe",
  "smart-keymap",
  "smart-keymap-nickel-helper",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -600,7 +600,6 @@ version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "41fea5ef5bed4d3468dfd44f5c9fa4cda8f54c86d4fb4ae683eacf9d39e2ea12"
 dependencies = [
- "defmt 0.3.100",
  "embassy-futures",
  "embassy-sync",
  "embassy-time",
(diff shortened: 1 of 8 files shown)


rgoulter:
Okay. I was able to get a onekey example working, using the smart keymap library, and the same `mod board` as the stm32f4-rtic `main.rs` uses. -- I was running into stack overflow problems, so needed to put the buffers & smart keymap backend into static memory.

It occurs to me, the use of `ld/stm32f4xx-tinyuf2/memory.x` isn't documented anywhere. I mean, my STM32F4 boards use tinyuf2 as the bootloader.

Also, I've not put `uf2conv` in the devenv.

Next step: I want to port over the MiniF4-36 examples from RTIC, demonstrating Uart with half-duplex.

rgoulter:
Hmm. Seems like `defmt` + smart keymap (with 36-key rgoulter) + split was a bit too much for the STM32F4xx MCU I have. -- I don't use defmt; but makes sense to have it as a feature for bins that do.

rgoulter:
I don't quite have the intuition for all the pieces regarding embassy & embedded & async & embassy-stm32's uart.

Seems that using blocking read doesn't work even in an async loop, since it blocks. I don't think using a spawned task would be any different, since STM32F4 is single core, AFAIU.

Using async Uart read is fine by me.

rgoulter:
I tried `[a_fut, b_fut, ...].iter().reduce(join)`, but the `_fut` types don't allow this. -- I think having so many `join`s smells.. but, might defer refactoring them to embassy tasks later.

rgoulter:
Current status:

- The LHS almost works. (works as a USB HID device, and matrix scanning works).
  - Not working: receiving messages over half-duplex UART.
- The RHS works. (Matrix scanning + sending messages over half-duplex UART).

(Tidying to do: `Cargo.toml` should conditionally support `defmt` as a `feature`; the LHS/RHS examples would surely benefit from using embassy tasks, rather than `join(join(join(...)))`).

rgoulter:
Hmm. This implementation of "reset to bootloader" involves an unsafe volatile write (magic value so that tinyuf2 enters bootloader). But, the smart keymap doesn't support setting the callback to use an unsafe function pointer.

Eh.

I think I'd rather keep the callbacks safe.

PR #303, opened 2025-04-13
usbd-smart-keyboard: allow access to keymap output

rgoulter:
Currently, the `KeyboardBackend` in `usbd-smart-keyboard` supports `usbd-human-interface-device`.

`embassy`'s USB is separate from `usb_device`. -- Still, `KeyboardBackend` seems like a useful abstraction.

This PR enhances `smart_keymap::keymap::KeymapOutput`, and adjusts `usbd_smart_keyboard::input::smart_keymap::KeyboardBackend` to allow references to the `KeymapOutput`.

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -30,7 +30,7 @@ const MAX_QUEUED_INPUT_EVENTS: usize = 32;
 pub const INPUT_QUEUE_TICK_DELAY: u8 = 1;
 
 /// Constructs an HID report or a sequence of key codes from the given sequence of [key::KeyOutput].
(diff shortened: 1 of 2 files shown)


rgoulter:
I *think* than `usbd-smart-keyboard` can be renamed to `keyberon-smart-keyboard`.

The `usbd-device`-specific stuff is: type aliases, common imports, and "usb poll". It's not really useful enough to have as its own crate. (Especially since the STM32F4 doesn't have enough USB EPs for USB serial; so, that part gets `#[cfg]`'d anyway).

PR #339, opened 2025-04-23
Callback Key: add KeymapCallback::Custom

rgoulter:
`KeymapCallback` was used to support invoking `Reset` and `ResetToBootloader` in keymap firmware.

Smart keyboards sometimes have extra functionality like backlighting, per-key RGB lighting, Bluetooth controls, etc. which can be modified with keys. e.g. "RGB_OFF" or "BLUETOOTH_PROFILE_2".

Adding a "Custom" variant to KeymapCallback at least allows this functionality. -- A keyboard firmware implementation can register callbacks for e.g. controlling RGB effect state, and the keymap can invoke those callbacks.


rgoulter:
FW size: 74.21%, up from 73.34%

PR #340, opened 2025-04-24
More Key outputs

rgoulter:
It's common for keyboard firmware to support key outputs other than keyboard codes. -- e.g. media keys (consumer), mouse keys, or maybe custom key codes.

I think many use cases for custom keys can be covered with the `Custom(i, j)` callback key; but, may be useful to support `Custom(i, j)` keys for arbitrary custom pressed key behaviour.


rgoulter:
77.0% -> 77.9%

PR #341, opened 2025-04-24
Tap Dance key

rgoulter:
Tap Dance!

I see that some other firmware (like fak, and ZMK) support using tap-hold keys as tap-dance definitions. That may be nice to support at some point.

The big blocker in implementing tap-dance was computing max # of tap-dance definitions. -- This requires traversing the key nodes in the keymap codegen. (Technically, layers should do the same thing; but, gets away with assuming at least one key is a layered key).

The `keymap.ncl` syntax for tap dances is `key & { tap_dances = [..] }`. Maybe it's a somewhat cumbersome way of defining them. But, if `tap_dance.make [..]` is preferred, Nickel supports writing functions like that. I like the emphasis on "smart key behaviour as a modifier of a typical key".


rgoulter:
74% -> 77%, currently.

(Though, this doesn't support tap-and-hold tap-dance definition, which might be a tiny bit more code).

PR #345, opened 2025-04-29
Add Custom Key

rgoulter:
Followup to #340.

This adds a key which outputs the `Custom` key usage.

I expect that the `Callback::Custom(i, j)` is going to be more useful for smart keyboard effects. But, there may be use cases where the keyboard firmware wants to know if a key is pressed or not.

I expect that the Custom key (which lets the user define keymaps with 256 custom codes, the behaviour of which is up to the keyboard firmware implementation).


rgoulter:
77.9% -> 78.2%.

PR #347, opened 2025-04-30
firmware: ch32: use mounriver riscv embedded gcc

rgoulter:
Hmm.

The code *builds* with the XPack disribution of GCC.

I thought it would also run.

However, I've been using a `build` configured to use mounriver's GCC, so never noticed that building with xpack, the ISRs don't seem to work.

(Fortunately, the keyboard can still enter the bootloader by holding SW_1_1).


rgoulter:
I'm interested in moving away from MRS' toolchain, & onto Rust. -- I'll have to try its ISR support.

rgoulter:
#348 adjusts the interrupt attribute params to work with xpack's riscv embedded gcc toolchain.

PR #348, opened 2025-05-01
firmware: ch32: remove "WCH-Interrupt-Fast" interrupt param

rgoulter:
I noticed that I'd been unintentionally building the CH32X firmware with the mounriver studio distribution of GCC, whereas the devenv is set to use the xpack distribution of GCC.

I noticed that the CH32X firmware built with the xpack gcc embedded toolchain doesn't work.

This seems to be  due to interrupt handling.

https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Function-Attributes.html

`"WCH-Interrupt-fast"` with the mounriver toolchain seems to make use of features in the CH32X MCU allowing faster interrupt handling.

Removing this attribute, the code works with xpack, and still works with MRS.

I see the CH58x code has this:

```
#ifndef __INTERRUPT
#ifdef INT_SOFT
#define __INTERRUPT   __attribute__((interrupt()))
#else
#define __INTERRUPT   __attribute__((interrupt("WCH-Interrupt-fast")))
#endif
#endif
```

I'm not sure where INT_SOFT comes from.

### DIFF
--- a/firmware/ch32x035-usb-device-compositekm-c/User/ch32x035_it.c
+++ b/firmware/ch32x035-usb-device-compositekm-c/User/ch32x035_it.c
@@ -10,8 +10,8 @@
 
 #include "ch32x035_it.h"
 
-void NMI_Handler(void) __attribute__((interrupt("WCH-Interrupt-fast")));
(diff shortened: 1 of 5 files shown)


rgoulter:
(Tried the CH58x firmware with xpack, and it did work).

PR #356, opened 2025-07-03
chorded tap-hold over tap-hold keys

rgoulter:
One use case which smart-keymap should be able to support: a chord which is a tap-hold key, where the chorded keys are themselves tap-hold keys.

This could be useful for making expressive use of tap-hold keys, such as tap-hold home-row modifiers, or tap-hold thumb keys.

### DIFF
--- a/src/key/chorded.rs
+++ b/src/key/chorded.rs
@@ -544,8 +544,11 @@ where
 
                     (pkr, pke)
                 }
-                // n.b. no need to add to key path; chorded aux_key only nests the passthrough key.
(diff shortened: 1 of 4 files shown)


rgoulter:
CNR bugs.

I was trying to chord each of the two tap-hold thumb keys as the third tap-hold thumb-key. With the CH32X-48, this .. doesn't work smoothly. (Hardware issue, even, as discussed in the keyboard-labs PR about the CH32X-48?). With the Pico42, I observe no problems.

PR #368, opened 2025-08-14
Overlapping chord support

rgoulter:


rgoulter:
Currently, smart-keymap has some support for partially overlapping chords.

This PR aims to iterate on the chording logic so that it correctly supports overlapping chords.

tests/rust/chorded/overlapping.rs
```diff
@@ -0,0 +1,164 @@
...
+));
+
+const CONTEXT: Ctx = key::composite::Context::from_config(composite::Config {
+    chorded: chorded::Config {
+        chords: [
+            Some(chorded::ChordIndices::from_slice(&[0, 1])),
+            Some(chorded::ChordIndices::from_slice(&[0, 1, 2])),
+            Some(chorded::ChordIndices::from_slice(&[1, 2])),
+            Some(chorded::ChordIndices::from_slice(&[2, 3])),
+        ],
```
rgoulter, line 67: n.b. In order to test against more than 4 chords, it'd be convenient to use the `Slice` for the `key::chorded::Config`, too.

tests/rust/chorded/overlapping.rs
```diff
@@ -0,0 +1,459 @@
...
+type AK = composite::ChordedKey<composite::Layered<composite::TapHold<keyboard::Key>>>;
+
+// 4-key keymap
+//   A B C D
+// chords:
+// 0: X X - - => M=0x10
+// 1: X X X - => N=0x11
+// 2: - X X - => O=0x12
+// 3: - - X X => P=0x13
+
```
rgoulter, line 26: I think it'd be worth adding "X - - X" as a chord, and checking "X - - X" vs "- - X X" as overlapping.

src/key/chorded.rs
```diff
@@ -102,18 +112,22 @@ impl Context {
...
 
-    /// Returns the chord indices for the given pressed indices.
-    ///
-    /// The returned vec is empty if any of the indices are not part of a chord.
-    pub fn chords_for_indices(
-        &self,
-        indices: &[u16],
-    ) -> heapless::Vec<ChordIndices, { MAX_CHORDS }> {
+    /// Returns the satisfiable chords for the given pressed indices.
+    pub fn chords_for_indices(&self, indices: &[u16]) -> heapless::Vec<ChordState, { MAX_CHORDS }> {
```
rgoulter, line 116: Hmm.

At the moment, this constructs which chords are satisfiable for the *given* indices.

For the "chord2" case, this is works.

But, for overlapping chords, this isn't quite the right abstraction.

Context keeps track of which indices are pressed. But, it should also keep track of resolved chords. -- One can imagine a keymap "ABCD", which chords "AB", "CD", and "ABCD". -- Want to be able to activate e.g. AB followed by (after a timeout) chord CD. So... `chords_for_indices` would better be replaced by a query for "chords for keymap index".

Perhaps something like:
- `[Chord { is_satisfied: true }]` if a chord is resolved,
- `[Chord { is_satisfied: true }, ..]` if an overlapped chord is satisfied.
- `[]` if no chords can be satisfied.

Then, in a situation like "ABCD", chords BC, CD; "press B,C,D" should resolve chord BC; and the query for keymap_index 3 should return [].

rgoulter:
Just noticed the implementation (even with Chord2/non-overlapping, although it's not really apparent in that implementation) makes mistaken assumptions about the event handling system with regards to PendingKeyState and Context.

The implementation was assuming that Context receives events alongside PendingKeyState; and so was relying on Context to keep track of state.

Instead, what happens is that when the Keymap has PKS, PKS exclusively receives events (& then subsequent events are replayed once the PKS resolves). So, PKS should consider cases from events, and resolve appropriately.

tests/rust/chorded/overlapping.rs
rgoulter: Another case to check: a chord with multiple aux keys.

And re-tapping one of the chorded keys. (e.g. "A B", with chord "A B": pressing AB, then releasing A & tapping A (or releasing B & tapping B).

src/key/chorded.rs
```diff
@@ -633,22 +778,29 @@ impl PendingKeyState {
...
                 // Timed out before chord unambiguously resolved.
-                //  So, the key behaves as the passthrough key.
-                Some(ChordResolution::Passthrough)
+                let maybe_satisfied_chord_id = self
+                    .satisfied_chord()
+                    .map(|chord_state| chord_state.index as u8);
+                match maybe_satisfied_chord_id {
+                    Some(satisfied_chord_id) => Some(ChordResolution::Chord(satisfied_chord_id)),
+                    _ => Some(ChordResolution::Passthrough),
+                }
```
rgoulter, line 787: Fairly sure clippy dislikes this.

rgoulter:
Tests all pass! That's positive.

The code needs some cleanup. Perhaps the reimplementation commit squashed into one. And, I'd feel more confident with those tests mentioned added.

PR #388, opened 2025-09-01
Refactor Rust Integration Tests to Use keymap! macro

rgoulter:


rgoulter:
This was a task I'd hoped the LLM agents would be good at. (e.g. asking the aider-chat program to manage the Gemini 2.5 Pro LLM). But, despite giving it the features.html documentation, it was unable to properly come up with the appropriate keymap ncl expressions.

PR #389, opened 2025-09-02
Indirection on new_pressed_key

rgoulter:
Towards #381, one blocker is that various `key::Key` implementations themselves invoke `key::Key::new_pressed_key`.

This PR is to rearrange the types so that these invocations instead return some kind of value.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -129,11 +129,34 @@ impl<E: Debug, const M: usize> IntoIterator for KeyEvents<E, M> {
     }
 }
 
+/// Newtype for invoking new_pressed_key on the key at the given [KeyPath].
+#[derive(Debug)]
+pub enum NewPressedKey {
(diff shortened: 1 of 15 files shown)


rgoulter:
Hmm.

Previously, when the implementation returns `KeyState` directly, it's able to return `NoOpKeyState`.

It's not obvious to me whether it's better to have a `NoOpKeyState`, or to have a `NewPressedKey::NoOp` variant. But, seems untidy to have both.

rgoulter:
Good news is, the test suite is quite comprehensive if it's catching issues.

rgoulter:
Regarding common code: it's true that lots of the code is common, especially in how the pending key state is handled in process_input and handle_event.. but, since both these methods mutably borrow self (`&mut self`), & since the code which processes the events with the pending state calls other methods on self which also mutably borrow self, the common code can't really be expressed as a method (since a method borrows at least `&self`).

PR #394, opened 2025-09-21
More Consistent key::composite Boilerplate

rgoulter:
I've been considering different ways of shifting the complexity of `key::composite` around.

`key::composite` aggregates all the other key implementations so that `keymap::Keymap` has access to all the keys. The logic in `key::composite` is very straightforward: it has sum types (enum of `Keyboard(x)`, `TapHold(x)`, ...) for Ref, Event, PendingKeyState, KeyState, and product types for Context, System. -- The logic is 'just' dispatching the methods of `key::System` appropriately.

If all the code in `key::composite`'s `key::system` implementation is dead simple to maintain, this reduces the maintenance burden of adding keys.

This PR adds more `From` impls, and changes the PKR/PKE uses to implicitly make use of these.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -588,13 +588,17 @@ impl<T: Copy> Event<T> {
     }
 
     /// Maps the Event into a new type.
-    pub fn try_into_key_event<U, E>(self, f: fn(T) -> Result<U, E>) -> EventResult<Event<U>> {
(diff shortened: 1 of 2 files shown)


rgoulter:
I figured Copilot's review would hate the `From` `panic!()`.

PR #402, opened 2025-09-21
ncl: modularize key-extensions.ncl

rgoulter:
This PR rewrites `key-extensions.ncl` so each `key_extension` is now a record. Access to other key fields is provided through its own `keys` field.

This has some benefits:

- The `key-extensions` are now easier to reason about. Functions are opaque, whereas records can be manipulated.
- Records can be described with nickel contracts.
- This makes it easier to auto-generate documentation.
- By relying on lazy evaluation, `keys` is not sensitive to the order of extensions.

One disadvantage is, as far as I could tell, was the `shifted` key extensions definition now needs to be written with more copy-pasting.

### DIFF
--- a/ncl/key-extensions.ncl
+++ b/ncl/key-extensions.ncl
@@ -1,5 +1,5 @@
 {
-  keyboard = fun K =>
+  keyboard =
     (import "hid-usage-keyboard.ncl")
     |> std.record.map_values (fun kc =>
       kc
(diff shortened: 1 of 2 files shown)


features/keymap/ncl/layer_string.feature
```diff
@@ -40,7 +40,8 @@ Feature: Layer String
     * a keymap.ncl:
       """
       {
-        custom_keys = fun K =>
+        custom_keys =
```
rgoulter, line 43: The markdow description above needs to be updated.

ncl/keymap-ncl-to-json.ncl
```diff
@@ -611,8 +611,8 @@ let validators = import "validators.ncl" in
   #   - String
   layer_as_array_of_keys = fun l =>
     if std.is_string l then
-      let { extend_keys, .. } = import "key-extensions.ncl" in
-      let K = extend_keys (import "keys.ncl") custom_keys in
+      # let { extend_keys, .. } = import "key-extensions.ncl" in
+      let K = (import "keys.ncl") & custom_keys in
```
rgoulter, line 615: `extend_keys` is good to keep. It'd allow extending with functions.

PR #416, opened 2025-09-25
Add Mouse keys

rgoulter:


rgoulter:
I'll note, this Nickel code is particularly bad:

```
          let { variant, rust_expr_val } =
            mouse_ref
            |> match {
              { Button = b } => { variant = "Button", rust_expr_val = std.to_string b },
              s if std.is_string s => { variant = s, rust_expr_val = "" },
              _ => std.fail_with "bad mouse_ref: %{std.serialize 'Json mouse_ref}",
            }
          in
          let rust_expr =
            if rust_expr_val == "" then
              "%{module}::Ref::%{variant}"
            else
              "%{module}::Ref::%{variant}(%{rust_expr_val})"
          in
```

It .. works, but it's a very silly way of writing it.

PR #429, opened 2025-09-27
Firmware: STM32-Embassy: add consumer, mouse support

rgoulter:
And, finally, following #428, #427, #420, #419 and others, this PR updates the STM32-embassy examples with consumer, mouse support.

As with the stm32F4-rtic examples.. note that STM32F4 has 4 endpoints, so control + 3 interfaces. So, not enough endpoints to also add e.g. serial, or other interfaces with this approach. (Could perhaps share the endpoint with a compound HID device).


stm32f4-rtic-smart-keyboard/src/main.rs
```diff
@@ -89,6 +89,7 @@ mod app {
         backend: KeyboardBackend,
         report_success: bool,
         timer: timer::CounterUs<pac::TIM3>,
+        previous_mouse: WheelMouseReport,
```
rgoulter, line 92: Uh. Let's just leave this file as-is.

PR #436, opened 2025-09-28
Replace use of `crate::init::MAX_` constants with generics for key::layered

rgoulter:
Some of the smart key implementations make use of pub consts from `crate::init`. (Layered, Tap Dance, Chorded).

This is awkward and 'dirty'.

This PR replaces use of these constants with generics in the `key::*` implementations, and moves use of these constants to `key::composite`. This is still arguably 'awkward', but the awkwardness is limited to `key::composite`.

### DIFF
--- a/src/key/composite.rs
+++ b/src/key/composite.rs
@@ -8,6 +8,8 @@ use serde::Deserialize;
 
 use crate::{key, keymap};
 
+use crate::init::LAYER_COUNT as LAYERED_LAYER_COUNT;
+
 /// Aggregate enum for key references.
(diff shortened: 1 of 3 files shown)


rgoulter:
Firmware size goes up a tiny bit?

rgoulter:
Ah, yup. This compromise is nice:

- Removing use of `crate::init::*` in the `key::*` implementations is less awkward.
- Using the `init::*` constants in `key::composite` is a practical compromise. `key::composite` and the `crate::init` codegen are already quite tightly coupled.

PR #523, opened 2026-02-20
SetActiveLayers: support layer mask

rgoulter:
## Summary

Adds **layer masking** to `SetActiveLayers`, so a layer-modifier key can set which layers are active **within a subset** of layers, without changing layers outside that subset.

This supports **orthogonal layer groups** in one keymap — for example, switching among "OS desktop" layers (Windows / Linux / macOS bindings) while LOWER, RAISE, or other unrelated layers stay as they were.

## Problem

`SetActiveLayers` previously replaced the full active-layer bitset. Picking one layer in a mutually exclusive group (e.g. "Windows desktop") would clear every other active layer, including layers used for unrelated purposes.

## Behaviour

`SetActiveLayers` now carries a `ModifierBitset { layers, mask }`:

- For each layer index in `mask`: activate if the bit is set in `layers`, otherwise deactivate.
- Layers **outside** `mask` are left unchanged.

Unmasked `set_active_layers_to` keeps the old semantics: `mask` defaults to all layers (`BITSET_MASK_ALL`), so existing keymaps are unchanged.

Hold, Toggle, and Sticky are unchanged — they already affect a single layer and do not replace the bitset.

## Authoring (`keymap.ncl`)

Existing API (full bitset replace):

```nickel
K.layer_mod.set_active_layers_to [1]
```

Masked set (new):

```nickel
K.layer_mod.set_active_layers_amongst {
  set_active_layers_to = [3],
  affected_layers = [3, 4, 5],
}
```

Only layers 3, 4, and 5 are updated; layer 1 (e.g. LOWER) remains active if it was already on.

## Implementation

| Area | Change |
|------|--------|
| Rust | `ModifierBitset`, masking in `LayerEvent::Set`, `set_active_layers_from_bitset_with_mask` |
| Nickel codegen / ncl-to-json | `{ layers, mask }` JSON for `SetActiveLayers` |
| `layered-key.ncl` | Layer-state sim for Cucumber input resolution |
| Tests | Rust integration tests, Cucumber feature, ncl contract check |

## Tests

- `tests/rust/layered/set_active_layers.rs` — preserve outside mask, switch within mask, unmasked replace
- `features/keymap/key/layer_modifier-set_active_layers_mask.feature`

## Related

Motivated by context-dependent / "semantic" keys built on `LayeredKey` (same slot, different output per active variant). Masked `SetActiveLayers` is the layer-switching primitive; Nickel authoring sugar can follow separately.

### DIFF
--- a/features/keymap/key/layer_modifier-set_active_layers_mask.feature
+++ b/features/keymap/key/layer_modifier-set_active_layers_mask.feature
@@ -1,0 +1,85 @@
+Feature: Layer Modifier: Set Active Layers Amongst (Mask)
+
+  The `K.layer_mod.set_active_layers_amongst` key sets active layers within a
+  mask, leaving layers outside the mask unchanged.
+
+  This supports orthogonal layer groups, e.g. switching "OS desktop" layers
+  without affecting unrelated layers such as LOWER or RAISE.
+
+  Background:
+
+    Given a keymap.ncl:
+      """
+      let K = import "keys.ncl" in
+      {
+        layers = [
+          [
+            K.layer_mod.set_active_layers_to [1],
+            K.layer_mod.set_active_layers_amongst {
+              set_active_layers_to = [3],
+              affected_layers = [3, 4],
+            },
+            K.layer_mod.set_active_layers_amongst {
+              set_active_layers_to = [4],
+              affected_layers = [3, 4],
+            },
+            K.A,
+            K.G,
+            K.D,
+          ],
+          [K.TTTT, K.TTTT, K.TTTT, K.TTTT, K.B, K.E],
+          [K.TTTT, K.TTTT, K.TTTT, K.TTTT, K.TTTT, K.TTTT],
+          [K.TTTT, K.TTTT, K.TTTT, K.TTTT, K.C, K.TTTT],
+          [K.TTTT, K.TTTT, K.TTTT, K.TTTT, K.F, K.TTTT],
+        ],
+      }
+      """
+
+  Example: masked set active layers preserves layers outside the mask
+
+    When the keymap registers the following input
+      """
+      [
+        tap (K.layer_mod.set_active_layers_to [1]),
+        tap_keymap_index 5,
+        tap (K.layer_mod.set_active_layers_amongst {
+          set_active_layers_to = [3],
+          affected_layers = [3, 4],
+        }),
+        tap_keymap_index 4,
+        tap_keymap_index 5,
+      ]
+      """
+    Then the output should be equivalent to output from
+      """
+      [
+        tap (K.E),
+        tap (K.C),
+        tap (K.E),
+      ]
+      """
+
+  Example: masked set active layers switches within the mask
+
+    When the keymap registers the following input
+      """
+      [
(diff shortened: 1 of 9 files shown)


rgoulter:
Hmm.

I think in order to get the "desktop keys" that I have, this would require "conditional layers" in addition to this "set layers (with mask)" functionality. -- Which then means each layered key is that much larger. (Albeit, this would be less of a problem with "sparse layered key").

Hmm.
I guess for the "desktop keys" (e.g. workspace left/right, for Linux/macOS/Windows), what I was thinking of was "activatable"? where "desktop" would activate all the desktop layers, but this 'activatable' would restrict it to one at a time? -- I think "conditional keys" sounds a broader solution.

rgoulter:
RE: Conditional Layers.

It's tempting to think about abstractions like a full-on Kirei-style "key expression". -- I think with careful design, that might be an interesting abstraction to try, but I'm not sure it's a good fit for smart keymap. (For example, in my mind: smart keymap keeps each key system separate (layered doesn't know about tap hold, etc.; whereas, querying would either depend on a common central system or, involve some kind of way to refer to state of other system?).
-- But on the other hand, what's *possible* and what's *practical*/*useful* are different for keymaps.

Conditional layers are a generalisation of tri-layer.

My motivating use case is 'desktop keys'; keys which behave differently depending on condition/context.

I recall some keyboards even have external DIP switches. Smart-keymap wouldn't really support these external conditions.

Perhaps what I want is: keymap has some kind of config registry. Each config register supports 32 bits (say). Perhaps at Keymap scope. (System-specific 'registers' can be implemented as contexts). So, this condition key would behave differently depending on how the register is set; like layers, it could support variants, but would have keys to manipulate the registers directly.

(Again, though: having a key that's similar to layers but not layers... smart-keymap smells like it has some redundancy in its key systems).

rgoulter:
Ah.

Another thought:
This "case key"/"condition key" that I want for "desktop keys" ~could likely borrow from the same layer system. But the obvious blocker is: right now adding a layer adds to the size of every layered Key, since it's all in the same "layer space". There's one "layer space", and all layered keys are activated with the same layer state.

If could have e.g. the "layers" layer state, "desktop" layer state, then the layered keys could borrow from this.

May-be also in miryoku style keymaps: the left hand layers are practically separate from the right hand layers. ("fallthrough"/"transparent" is practically the same as "not affected by that layer state"). -- Though I suspect this gets tricky: e.g. if there are base layers (qwerty, dvorak, ...), then it's not as simple as "left hand layer state, right hand layer state".

Hmm.
Perhaps simpler to think of "layer space" as "subset of layers". Ah. So rather than `[Key]`, if instead it's `{ LayerSet, [Key] }` then for a bit of additional overhead (computation and storage), layers can be stored more compactly. -- That way, having "Win layer", "macOS layer", "Linux layer" would only involve the overhead of may be having a larger layerset. (Presumably a bitset).

-- In which case, *then* having "Set Active Layers" have a layerset mask makes sense. And perhaps it makes sense for other operations to operate using a layerset, too. For "desktop keys", these could be just implemented in keymap-ncl-to-json as additional layers.

PR #527, opened 2026-05-23
ncl: strict key extensions evaluation

rgoulter:
Avoid errors related to infinite recursion, which somehow regressed from earlier `nickel` versions.

PR #530, opened 2026-05-24
Fix nickel infinite recursion

rgoulter:
Alternative to #527, fixing the infinite recursion.


rgoulter:
Hmm. This fails locally but not in CI? Oh, CI downloads Nickel specifically.

PR #536, opened 2026-05-24
Fix ci gendoc

rgoulter:


rgoulter:
Oh, another MR un-split the tests.

Again, though: previously it took 3m, now it takes 19m!

PR #548, opened 2026-07-04
ncl: discover snapshot fixtures; use cargo check in compile tests

rgoulter:
## Summary

Simplifies the NCL snapshot test harness: compile verification only needs `cargo check`, and fixture directories are discovered from the filesystem instead of a hand-maintained list.

### `cargo check` instead of `cargo rustc`

`test-ncl-builds.sh` verifies that codegen'd `keymap.rs` type-checks for the firmware target (`riscv32imac-unknown-none-elf`). Linking release artifacts for each of 16 fixtures is unnecessary; `cargo check` is sufficient and faster.

### Auto-discover snapshot fixtures

New `list-ncl-snapshot-fixtures.sh` lists `tests/ncl/keymap-*` directories that contain `expected.rs` (sorted, one per line).

`run-tests.sh` and `save-test-snapshots.sh` use it via `mapfile` — no duplicated hard-coded arrays. Adding a snapshot fixture only requires committing `expected.rs`; removing the snapshot drops the fixture from the suite automatically.

Fixtures without `expected.rs` (e.g. `keymap-36key-miryoku`) are intentionally excluded.

## Test plan

- [x] `make test-ncl`

### DIFF
--- a/ncl/scripts/list-ncl-snapshot-fixtures.sh
+++ b/ncl/scripts/list-ncl-snapshot-fixtures.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+# Lists tests/ncl fixture directories that have expected.rs snapshots.
+# One fixture name per line, sorted for stable ordering.
+
+set -euo pipefail
+
+SCRIPTS_DIR="$(dirname "$0")"
+NCL_TESTS_DIR="${SCRIPTS_DIR}/../../tests/ncl"
+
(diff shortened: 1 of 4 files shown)


rgoulter:
Split into #549 and #550 for atomic PRs.

rgoulter:
Split into #550 (fixture discovery) and #549 (cargo check, stacked on #550).

PR #352, opened 2025-05-03
Nickel: improve UX of the KeymapLayerString contract

rgoulter:
Previously, #351 added the contract `KeymapLayerString`, but just as an alias for the `String` contract.

The errors from evaluating malformed `keymap.ncl` can be unclear.

I think the validators/contracts ought to be useful in at least the following cases:
- for `[Key]` layer, if some non-`Key` is provided.
- for `KeymapLayerString` layer, if some value isn't a field in `keys`/`extended_keys`.

PR #556, opened 2026-07-05
ncl: contract layered_keys and chorded_keys pipeline outputs

rgoulter:
Continues the #352 / #551–#554 work on Nickel error messages.

Those PRs improved validation at **author input** (`layers`).

This PR closes the next gap: **pipeline outputs** that were still unchecked until deep inside `json_keymap` evaluation.

## The problem: two phases, one contracted

`keymap-ncl-to-json.sh` exports `--field=json_keymap`. Nickel evaluates a transform pipeline. Contracts only ran where fields were annotated — mostly `layers` (and config like `chords`). Everything below was plain evaluation: a bad derived key could sail through layer validation and fail dozens of frames later (i.e. a long/unclear error message), or only when `to_json_value` hit the wrong shape.

```
keymap.ncl  (author writes)
  │
  │  layers ──contract──► KeymapLayer          ✅ #551–#554
  │     │                    (string tokens or Array KeymapKey)
  │     ▼
  │  layer_as_array_of_keys                    (resolve strings → records)
  │     ▼
  │  layered_keys                              ❌ was uncontracted
  │     │                    merge layers → plain key, or
  │     │                    base_key & { layered = [ … ] }
  │     ▼
  │  chorded_keys                              ❌ was uncontracted
  │     │                    plain key, { passthrough = k }, or
  │     │                    { chords, passthrough }
  │     ▼
  └  json_keymap                               automation fold + to_json_value
```

So we had good errors for “unknown token in layer string” and “bad key in `layers = [ … ]`”, but not for “this **merged layered composite** is malformed” or “this **chord wrapper** has an invalid inner key” — failures that only appear after transforms run.

## Why not `Array KeymapKey` on pipeline outputs?

`KeymapKey` is for **author-facing** input:

- `layers = " A B "` / `layers = [ K.A, K.B ]`
- `keys = [ K.A, K.B ]` directly
- explicit composites in source: `K.Tab & K.hold … & { layered = [ … ] }`

Pipeline slots are **stage-specific shapes**:

| Field | After transform, each slot is… |
|-------|--------------------------------|
| `layered_keys` | A plain `KeymapKey`, **or** `{ layered, ..base }` from multi-layer merge |
| `chorded_keys` | A plain key, **`{ passthrough = k }`** (auxiliary), or **`{ chords, passthrough }`** (primary) |

Using `Array KeymapKey` on those outputs would re-use export dispatch semantics for validation — 'not incorrect', but the wrong abstraction.

Stage contracts document what each transform may emit.

## What this PR adds

Explicit validators (field-presence branching, not `key_modules` first-match):

- **`LayeredKeyElement`**: `layered` field → `layered.Key`; else → `KeymapKey`
- **`ChordedKeyElement`**: `chords` → `chorded.Key`; `passthrough` → `chorded_aux.Key`; else → `KeymapKey`

Attached as:

```nickel
layered_keys | Array LayeredKeyElement = …
chorded_keys | Array ChordedKeyElement = …
```

Nickel now checks these **results** before `json_keymap` consumes them — so a bad composite fails at the pipeline stage that produced it, with a contract error instead of a deep eval stack.

Checks: `checks.check_layered_key_element`, `checks.check_chorded_key_element` (plain keys, composites, invalid inner passthrough).

## Relationship to #555

#555 made `key_modules` order the spec for **export** when one record matches several modules (e.g. tap-hold thumb + `layered` stack). #556 is orthogonal: it names and validates **what the layering/chording transforms output**. Export still uses `key_modules`; pipeline contracts use explicit branches.

## Not in scope

- `json_keymap` automation transform output (next gap if we want the same treatment)
- Tightening author `KeymapKey` / `KeymapLayer` — intentionally still loose

## Testing

- `make test-ncl-checks` — includes new element validators and #555 dispatch checks
- `json_keymap` export on rgoulter / seniply / chorded fixtures

### DIFF
--- a/ncl/keymap-ncl-to-json.ncl
+++ b/ncl/keymap-ncl-to-json.ncl
@@ -307,6 +307,12 @@
       check_plain_key_ok =
         layered_key_element_validator K.A == 'Ok,
 
+      check_null_ok =
+        layered_key_element_validator null == 'Ok,
+
+      check_string_token_ok =
+        layered_key_element_validator K.caps_word.toggle == 'Ok,
+
       check_layered_composite_ok =
         layered_key_element_validator (K.A & { layered = [ null, K.B] }) == 'Ok,
 
@@ -319,6 +325,12 @@
     {
       check_plain_key_ok =
         chorded_key_element_validator K.A == 'Ok,
+
+      check_null_ok =
+        chorded_key_element_validator null == 'Ok,
+
+      check_string_token_ok =
+        chorded_key_element_validator K.caps_word.toggle == 'Ok,
 
       check_passthrough_ok =
         chorded_key_element_validator { passthrough = K.A } == 'Ok,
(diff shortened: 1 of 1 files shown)


rgoulter:
Earlier, I'd thought LLMs made for a good "first draft".

LLMs can quickly get you something, and something is better than nothing.

That's kinda true with PR descriptions.

I had thought: "oh, commit messages and PR descriptions must be human written". But it's more than they should have 'human thought' put into some distinct part of it. -- I want the PR to be 'human readable'.

In any case, LLMs can still do a good 'first draft', but really benefit from refining what's really intended.

rgoulter:
Also, for example, thinking about "LLM writes the code":

```

  chorded_key_element_validator = fun k =>
    if std.is_record k then
      if std.record.has_field "chords" k then
        keymap_ncl.chorded.key_validator k
      else if std.record.has_field "passthrough" k then
        keymap_ncl.chorded_aux.key_validator k
      else
        keymap_ncl.nullable_key.key_validator k
    else
      keymap_ncl.nullable_key.key_validator k,
``` 

This code "isn't wrong".

Some people might even like this code.

I do not.

It's difficult to keep track of what's 'true' on each line. A `match` expression is .. technically equivalent, but 'simpler'. (It's not nested).

As in ... it's LLM that writes the bytes, but I'm still keeping an eye on code style (for better or worse).

PR #558, opened 2026-07-05
ncl: propagate module validator messages for undispatched keys

rgoulter:
Continues the #352 / #551–#556 error-message work. After layer and pipeline contracts, wrong **key shapes** often still surfaced a generic fallback even when a smart-key module had a specific validator message.

## Problem

`key_validator` dispatches via `key_modules` + `is_key`. When nothing matches (e.g. `K.caps_word` record instead of `K.caps_word.toggle` string), export reported:

```
Value is not a valid keymap key
```

…while `caps_word.key_validator` already knows the fix: `Expected "ToggleCapsWord"`.

`is_key` is defined as `'Ok == key_validator`, so non-matching shapes never reach the module validator through normal dispatch.

## Before / after

`keymap.ncl`:

```nickel
let K = import "keys.ncl" in {
  keys = [ K.caps_word ],  # meant K.caps_word.toggle
}
```

`nickel export … keymap.ncl --field=json_keymap`

**Before (master):**

```
error: contract broken by the value of `layers`
       Value is not a valid keymap key
   ┌─ ncl/keymap-ncl-to-json.ncl:86:13
   │
86 │     | Array KeymapLayer
   │             ----------- expected array element type
   │
   ┌─ keymap.ncl:1:41
   │
 1 │ let K = import "keys.ncl" in { keys = [ K.caps_word ] }
   │                                         ^^^^^^^^^^^ applied to this expression
```

**After (#558):**

```
error: contract broken by the value of `layers`
       Expected "ToggleCapsWord"
   ┌─ ncl/keymap-ncl-to-json.ncl:86:13
   │
86 │     | Array KeymapLayer
   │             ----------- expected array element type
   │
   ┌─ keymap.ncl:1:41
   │
 1 │ let K = import "keys.ncl" in { keys = [ K.caps_word ] }
   │                                         ^^^^^^^^^^^ applied to this expression
```

Same blame location; message now comes from the caps_word module validator via shape hints.

## Approach

When dispatch finds no matching module:

1. **`key_validator_hint_modules`** — record-shape `match` hints at author intent (`{ toggle, .. }` → caps_word, `{ hold, .. }` → tap_hold, `{ key_code, .. }` → keyboard, chorded/layered/automation fields, etc.)
2. Run hinted module validators; return first error via **`validators.first_error_result`**
3. If no shape hints (e.g. `{ not_a_key = true }`), keep the generic message — avoids misleading caps_word text on arbitrary records

Dispatch order (#555) and pipeline stage contracts (#556) unchanged.

String-layer errors (`key HoldTab: …`) and per-element array blame (#553–#554) are unchanged.

## Testing

- `make test-ncl-checks` — `checks.check_key_validator_hints`, `validators.first_error_result`

### DIFF
--- a/ncl/keymap-ncl-to-json.ncl
+++ b/ncl/keymap-ncl-to-json.ncl
@@ -189,15 +189,58 @@
 
     Key = std.contract.from_validator key_validator,
 
+    # When no module's is_key matches, try validators for modules whose
+    # shape hints at author intent (e.g. K.caps_word record → caps_word).
+    key_validator_hint_modules = fun k =>
+      k
+      |> match {
+        { chords, .. } => [keymap_ncl.chorded],
+        { passthrough, .. } => [keymap_ncl.chorded_aux],
+        { layered, .. } => [keymap_ncl.layered],
+        { automation_instructions, .. } => [keymap_ncl.automation],
+        { tap_dances, .. } => [keymap_ncl.tap_dance],
+        { hold, .. } => [keymap_ncl.tap_hold],
+        { layer_modifier, .. } => [keymap_ncl.layer_modifier],
+        { keymap_callback, .. } => [keymap_ncl.callback],
+        { custom, .. } => [keymap_ncl.custom],
+        { toggle, .. } => [keymap_ncl.caps_word],
+        { consumer_code, .. } => [keymap_ncl.consumer],
+        { mouse, .. } => [keymap_ncl.mouse],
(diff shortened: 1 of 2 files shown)


rgoulter:
Closing for now — undispatched shape-hint approach may be revisited later; caps_word-style mistake-catching not worth the central registry / per-module hint_is_key overhead at this stage.

PR #570, opened 2026-07-05
ncl: OS semantic keys via named layers

rgoulter:
## Summary

Reimplements the OS desktop / semantic key sugar from the earlier draft **on top of named layers** (#594), instead of hard-coded layer indices 3/4/5.

No new Rust `key::*` types. Everything lowers to existing `LayeredKey` + masked `SetActiveLayers`.

## Nickel API

| Helper | Purpose |
|--------|---------|
| `K.semantic { windows = …, linux = …, macos = … }` | Named-layer `LayeredKey` (`{ layered = variants }`) |
| `K.os_windows` / `K.os_linux` / `K.os_macos` | Masked variant switch keys using **layer names** |
| `K.layer_mod.set_semantic_variant_to` | Generic named-variant switch for any name group |
| `K.os_desktop_layers` | `["linux", "macos", "windows"]` (alpha order) |

Layer names in `layer_mod` / OS switches are resolved to 1-based indices when compiling the keymap (after numbered layers, alphabetical named order).

Example:

```nickel
let desktop_left =
  K.semantic {
    windows = K.Left & K.LGUI & K.LeftCtrl,
    linux   = K.Left & K.LeftCtrl & K.LeftAlt,
    macos   = K.Left & K.LeftCtrl,
  }
in
{
  keys = [ K.os_windows, K.os_linux, K.os_macos, desktop_left ],
}
```

## Stack

Depends on #594 (named layers). Base branch: `ncl-named-layers`.

## Also in this PR

- `prepare_keys_named_layers`: lower named records + resolve string layer indices in `layer_mod`
- `inputs-to-json`: use prepared keys for lookup / layer state; fix `ReleaseKeymapIndex` emitting `Press`
- Cucumber: `features/keymap/key/semantic_os_desktop.feature`

## How to verify

```bash
ncl/scripts/run-ncl-checks.sh
cargo test -p smart-keymap --test cucumber-keymap -- -i '**/semantic_os_desktop.feature'
```

### DIFF
--- a/ncl/keymap-ncl-to-json.ncl
+++ b/ncl/keymap-ncl-to-json.ncl
@@ -1088,6 +1088,47 @@
             },
           ],
         },
+
+      # Same prepare path via authoring sugar (semantic + set_semantic_variant_to).
+      check_prepare_semantic_sugar =
+        let names = ["fn", "nav", "sym"] in
+        let keys = [
+          K.layer_mod.set_semantic_variant_to names "sym",
+          K.layer_mod.set_semantic_variant_to names "fn",
+          K.semantic { sym = K.A, fn = K.B, nav = K.C },
+        ]
+        in
+        {
+          actual = prepare_keys_named_layers keys |> std.array.map keymap_ncl.key.to_json_value,
+          expected = [
+            {
+              SetActiveLayers = {
+                layers = std.number.pow 2 3,
+                mask =
+                  (std.number.pow 2 1)
+                  + (std.number.pow 2 2)
+                  + (std.number.pow 2 3),
+              },
+            },
+            {
+              SetActiveLayers = {
+                layers = std.number.pow 2 1,
+                mask =
+                  (std.number.pow 2 1)
+                  + (std.number.pow 2 2)
+                  + (std.number.pow 2 3),
+              },
+            },
+            {
+              base = { key_code = 0 },
+              layered = [
+                { key_code = 5 },
+                { key_code = 6 },
+                { key_code = 4 },
+              ],
+            },
+          ],
+        },
     },
 
   chorded_keys
--- a/ncl/smart_keys/layered/keymap-ncl-to-json.ncl
+++ b/ncl/smart_keys/layered/keymap-ncl-to-json.ncl
@@ -72,6 +72,22 @@
             affected_layers = ["fn", "nav"],
           }
         ),
+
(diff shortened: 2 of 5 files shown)


ncl/smart_keys/layered/keymap-ncl-to-json.ncl
```diff
@@ -183,6 +183,71 @@
...
+          base = { key_code = 0 },
+          layered = [
+            null,
+            null,
+            { key_code = 4 },
+            { key_code = 5 },
+            { key_code = 6 },
+          ],
+        },
+      },
```
rgoulter: Yeah ... there's *no* chance this code works. I'm surprised the LLM coding agent thinks it could.

rgoulter:
I'll keep this branch open as WIP.

The LLM thinks it meets the requirement.. but, then gets stuck. (Unfortunately, the cucumber tests are quite slow, and don't really allow for rapid iteration).

I ... think this will either just need actual thought put into how to achieve this (I agree that changes to the Rust core lib aren't required), or if driving an LLM, very close monitoring.

rgoulter:
Thinking about this a bit more:

I can agree with the assessment it's "just" layered keys.

What's missing is "named layers" or "symbolic layers".

This would work if:
- km ncl to json could take in LK definition with a key like `{ named_layered_keys = { 'x => K.A, 'y => K.y } }`, gather all these layer names (Nickel tags), and transform them to layered key definitions.
- the Rust core lib already supports nested layered keys. (e.g. a key on layer 3 which refs a key on layer 8), so I think that'd otherwise work.
- Further (but maybe not scoped for *this*) would be a `named_layers` in km.ncl like `{ 'num = [ N.N0, K.N1, ... ] }` (I mean, a full layer).

ncl/keymap-ncl-to-json.ncl
```diff
@@ -697,6 +697,80 @@
...
+  resolve_layer_index = fun name_to_idx idx_or_name =>
+    if std.is_number idx_or_name then
+      idx_or_name
+    else if std.is_string idx_or_name then
+      if std.record.has_field idx_or_name name_to_idx then
+        std.record.get idx_or_name name_to_idx
+      else
+        std.fail_with "unknown named layer: %{idx_or_name}"
+    else
+      std.fail_with "expected layer index number or name string",
```
rgoulter: Prefer a `match` block on the type.

PR #578, opened 2026-07-06
rust: keymap: fix scheduled input double-processing during pending

rgoulter:
## Summary

Fixes a bug where `Event::Input` delivered via `handle_event` during pending key state could be applied twice: `update_pending_state` at the top of `handle_event`, then again via `process_input` — and if the first pass resolved pending, `process_input` fell through to the normal press path and spuriously pressed the interrupting key.

Relevant for home-row tap-hold mods with `HoldOnKeyPress` interrupt behaviour.

## Changes

- Add `record_pending_input` (queue only)
- `handle_input` → `process_input`: record + apply (unchanged)
- `handle_event` + `Input` while pending was active: record for replay only if still pending; skip `process_input` after resolve

## Tests

- `scheduled_input_during_pending_does_not_reprocess_as_physical_press` — repro (now passes)
- `physical_input_during_pending_records_once_in_queued_events` — control

```
cargo test -p smart-keymap --lib
```

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -782,6 +782,30 @@ impl<
 }
 
 #[cfg(test)]
+impl<
+        I: Debug + Index<usize, Output = R>,
+        R: Copy + Debug,
+        Ctx: Debug + key::Context<Event = Ev> + SetKeymapContext,
+        Ev: Copy + Debug,
+        PKS: Debug,
+        KS: Copy + Debug + From<key::NoOpKeyState>,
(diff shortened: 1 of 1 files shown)


rgoulter:
Oh. Wow. This has been a very annoying bug!

And the keymap input pending handling was really thorny and complicated.

I'm glad the coding agent was able to identify this. And, surprisingly, each of the PRs to fix it have been relatively simple.

PR #594, opened 2026-07-10
ncl: support named layers on layered keys

rgoulter:
## Summary

Adds **named layers** as a first-class form of `layered` on keys in `keymap.ncl` → JSON.

### Numbered vs named

```nickel
# Numbered (existing): array slots → layers 1..N
K.A & { layered = [null, K.B, K.C] }

# Named (new): name → key record → layers after numbered
K.NO & { layered = { linux = K.PageUp, macos = K.Left } }
```

### Dense layout (what JSON / codegen see)

```
base
+ numbered layers 1..N     (array-form `layered`)
+ named layers   N+1..N+M  (record-form names, alphabetical)
```

No reserved gap before named layers. If there are no numbered layers, named layers start at layer **1**. (Earlier semantic-key sketches that hard-coded windows=3 / linux=4 / macos=5 were the wrong model.)

Example with both:

```nickel
keys = [
  K.D & { layered = [null, K.E] },   # N = 2
  K.NO & { layered = { windows = K.A, linux = K.B, macos = K.C } },
]
```

Named names sorted: `linux`, `macos`, `windows` → layers **3, 4, 5**.

### Lowering pass

Authoring keeps the record. One pass in `json_keymap` materialises dense arrays so every key shares the same global name → index assignment (including keys that only define a subset of the names).

**Skipped entirely** when no key uses the record form — no extra walk for ordinary keymaps.

Pipeline:

1. Collect named-layer field names across keys (nested via `map_accum`)
2. Sort alphabetically (Nickel `std.record.fields`)
3. Lower each layered key: numbered slots, then named slots (`null` where undefined)
4. Pad to global max length (existing behaviour)
5. `to_json_value` (array form only)

### Exported helpers

| Field | Meaning |
|-------|---------|
| `named_layer_names` | Sorted name list |
| `named_layer_indices` | name → 1-based layer index (for `layer_mod.*`) |
| `numbered_layer_count` | Max array-form length N |
| `max_layered_length` | N + M |

### Follow-ups (not in this PR)

- Keymap-level control of named-layer order (override alpha)
- OS-desktop / semantic sugar (#570) should build on **names**, not hard-coded indices
- Optional: `layer_mod` accepting a layer name string

## Test plan

- [x] `ncl/scripts/run-ncl-checks.sh` (named-layer checks + existing)
- [x] Mixed numbered + named JSON export
- [x] Lowering no-op when only array-form layered keys
- [x] Rebased on current `origin/master`

### DIFF
--- a/ncl/keymap-ncl-to-json.ncl
+++ b/ncl/keymap-ncl-to-json.ncl
@@ -1,3 +1,52 @@
+# keymap.ncl -> keymap.json
+#
+# This module is a multi-pass transform:
+#  author-facing fields are gradually rewritten
+#  into the dense JSON shape consumed by codegen / firmware.
+#
+# Pipeline (authored -> intermediate -> output)
+# -------------------------------------------
+#
+# - Authored inputs (keymap.ncl surface)
+#   - `keys` / `layers`   KeymapLayer: string tokens or Array Key
+#   - `chords`            [{ indices, key }, …]
+#   - `config`            optional per-feature config
+#   - `custom_keys`       extends token vocabulary for string layers
+#   Per-key sugar still present here:
+#     { layered = [Key|null, ...] }            # numbered layers (index i -> layer i+1)
+#     { named_layers = { name = Key|null } }   # named layers (lowered later)
+#     automation keys with instruction lists (e.g. string macros)
+#     ordinary composite keys (tap_hold, ...)
+#
+# - Intermediate key arrays (same length; each pass rewrites slots)
+#   - `layers_of_keys`     layers normalized to Array Key
+#   - `layered_keys`       multi-layer authoring folded into base & { layered }
+#                           (per-key `named_layers` still allowed)
+#   - `chorded_keys`       global `chords` attached as chorded / passthrough
+#   - `automation_transform`
+#       { automation_instructions, keys }
+#       macros flattened: keys hold { start, length } refs into the instruction list
+#
+# - Late key rewrites (inside `json_keymap`, after automation)
+#   - lower named layers
+#       named_layers -> dense `layered` array layout: [numbered slots…] ++ [named slots…]
+#       global name  -> index is alphabetical (see named_layer_*)
+#   - pad layered arrays   every layered key shares the same array length
+#   - to_json_value        per-key module projection (array-form layered only)
+#
+# - Output (`json_keymap`)
+#   {
+#     config = author config
+#              & { automation.instructions? }
+#              & { chorded.chords = indices },
+#     keys   = [ json key, … ],
+#   }
+#
+# Exported layer-index helpers (for layer_mod / docs; 1-based like layered[i]):
+#    numbered_layer_count, named_layer_names, named_layer_indices, max_layered_length
+#
+# Nesting: several passes walk composite keys via keymap_ncl.*.map_accum so
+# layered / named_layers / automation under tap_hold etc. are rewritten too.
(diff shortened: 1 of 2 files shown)


ncl/smart_keys/layered/keymap-ncl-to-json.ncl
```diff
@@ -207,33 +229,60 @@
...
+          let names = std.record.fields layered_ in
+          let { acc, layered } =
+            names
+            |> std.array.fold_left
+              (fun { acc, layered = parts } name =>
+                let child = std.record.get name layered_ in
+                let { acc, k = new_child } = f acc child in
+                {
+                  include acc,
+                  layered = parts & { "%{name}" = new_child },
```
rgoulter: This ... isn't obvious.

ncl/keymap-ncl-to-json.ncl
```diff
@@ -483,21 +483,24 @@
           )
           (std.array.length base_keys),
 
-  # Longest `layered` array on any layered key in the tree (not only the first key).
-  # Nested layered keys under composites are found via map_accum.
-  layered_length_of = fun k =>
+  # Numbered (array-form) layered length of a layered key; 0 for named-record form.
```
rgoulter: I think this "needs" an explanation as to why "named layers" get excluded.

I'd think e.g. if a KM has N numbered layers (excl. base) and M named layers, then the "LK" with a named layer would have (say) N null, with up to M additional keys above that.

ncl/smart_keys/layered/keymap-ncl-to-json.ncl
```diff
@@ -194,50 +205,144 @@
...
+        let without_layered =
+          if std.record.has_field "layered" k then
+            let { layered = _, ..rest } = k in rest
+          else
+            k
+        in
+        if std.record.has_field "named_layers" without_layered then
+          let { named_layers = _, ..rest } = without_layered in rest
+        else
+          without_layered,
```
rgoulter: Hmm. ... This smells unidiomatic.

ncl/smart_keys/layered/keymap-ncl-to-json.ncl
```diff
@@ -194,50 +205,126 @@
...
+        results
+        |> std.array.fold_left
+          (fun acc v =>
+            acc
+            |> match {
+              'Ok => v,
+              err => err,
+            }
+          )
+          'Ok,
```
rgoulter: Looks useful; should be in validators.ncl, or?

ncl/keymap-ncl-to-json.ncl
```diff
@@ -507,24 +562,152 @@
...
-          layered
-        else
-          layered @ std.array.generate (std.function.const null) (target_len - len)
-      in
-      base_key & { layered = padded_layered }
-    else
-      k,
+    k
+    |> match {
+      k if !(std.is_record k && keymap_ncl.layered.is_key k) => k,
```
rgoulter: Not quite.

rgoulter:
Retrospect:

- The LLM coding agent did enough of the work here that I give it credit.
- But... the LLM coding agent also displayed 'poor taste' beyond that point.
  - e.g. its code *worked*, but used `if`/`else` rather than pattern matching.
  - Or even in pattern matching, it might prefer `if`/`else` anyway!
- This PR took a lot of reworking to get to a good point.

PR #595, opened 2026-07-10
feat: keyboard modifiers on consumer and mouse keys

rgoulter:
## Summary

- Allow HID keyboard modifiers (e.g. Left Shift, Left Ctrl) on consumer and mouse base keys via the same Nickel merge as keyboard keys: `K.VolumeUp & K.LeftShift`, `K.MouseBtn1 & K.LeftCtrl`.
- While held, keyboard report carries the modifiers and consumer/mouse reports carry the leaf usage; release clears both.
- Documents the assessment and rejects a generic `KbdMod<R>` wrapper in `sketch-kbd-mod-base-keys.md` (same rationale as layer Hold-only mods).

## Changes

- **Rust:** `consumer::Ref::UsageCodeAndModifiers`, `mouse::Ref::ActionAndModifiers` + `Action`; `key_output` uses `from_usage_with_modifiers`.
- **Nickel:** optional `modifiers` on consumer/mouse validators, JSON, and codegen.
- **Tests:** rust integration + cucumber features `consumer-modified` / `mouse-modified`.

## Test plan

- [x] `cargo test --lib -- key::consumer key::mouse key::composite`
- [x] `ncl/scripts/run-ncl-checks.sh`
- [x] `cargo test --test rust-integration -- consumer mouse`
- [x] `cargo test --test cucumber-keymap -- --name modified`
- [ ] CI green on PR


src/key/mouse.rs
```diff
@@ -25,6 +48,34 @@ pub enum Ref {
     WheelLeft,
     /// Scroll wheel right.
     WheelRight,
+    /// A mouse action with keyboard modifiers. (action, modifiers_byte).
+    ActionAndModifiers(Action, u8),
```
rgoulter: Hmm.

No. This smells wrong. -- I mean the tests work, but this variant is clearly intented to be composite. This just smells like the wrong way to approach it.

rgoulter: e.g. the variants of `Ref` have been factored out to `Action`, but then duplicated in `Ref`?

Probably should just be `Action(Action)` variant? Or a `{ Action, mod }` Product?

src/key/consumer.rs
```diff
@@ -9,6 +9,8 @@ use crate::key;
 pub enum Ref {
     /// A usage code. (Value is the HID usage code).
     UsageCode(u8),
+    /// A usage code with keyboard modifiers. (usage_code, modifiers_byte).
+    UsageCodeAndModifiers(u8, u8),
```
rgoulter: Eh. Better to use a struct variant than a tuple variant here?

sketch-kbd-mod-base-keys.md
```diff
@@ -0,0 +1,326 @@
+# Sketch: Keyboard modifiers on base (leaf) keys
```
rgoulter: >.< Let's not check this in.

src/key/mouse.rs
```diff
@@ -27,6 +27,31 @@ pub enum Ref {
...
 
+/// Reference for a mouse key: an [Action] with optional keyboard modifiers.
+#[derive(Deserialize, Debug, Clone, Copy, PartialEq)]
+pub struct Ref {
+    /// The mouse action.
+    pub action: Action,
+    /// Keyboard modifiers byte ([key::KeyboardModifiers]).
+    #[serde(default)]
+    pub modifiers: u8,
+}
```
rgoulter: Not quite.

Technically, the intent is that `Ref` should be VERY light. (Whereas, this increases the sizeof `Ref`)

smart-keymap-core/src/key/mouse.rs
```diff
@@ -25,6 +53,35 @@ pub enum Ref {
     WheelLeft,
     /// Scroll wheel right.
     WheelRight,
+    /// Index into the key data array of [System] for a [Key] (action + modifiers).
+    Key(u8),
```
rgoulter, line 42: Still silly.

PR #597, opened 2026-07-10
test: expand pending delay-line vs session-log coverage

rgoulter:
## Summary

Expands unit test coverage for the delay-line vs session-log interactions around pending key resolution, as recommended in `sketch-pending-event-resolution.md` before another pending-local ingest pacing attempt.

These tests pin contracts that were easy to break in the 2026-07-08 pacing spike (when replays run relative to the delay line, what is and is not in the session log, and how the delay counter differs after resolve-via-tick vs resolve-inside-`handle_input`).

Keymap pacing tests are structured as focused AAA unit tests (one action / one property each), not multi-phase sagas. They stay under `#[cfg(test)]` rather than `tests/rust` because they inspect white-box queue/session-log state; `ObservedKeymap` auto-ticks and only exposes HID reports.

### Test hooks

- `test_is_pending`, `test_input_queue_len`, `test_input_queue_delay`
- `test_pending_session_log_inputs` (session-log `Event::Input`s only)
- `InputEventQueue::delay_counter()`

### Coverage

**Replay policy** (`pending.rs`):

- Resolve: filtered session-log inputs prepended *before* existing delay-line tail
- Resolve: non-input events scheduled with staggered delays
- Nested pending: LIFO input prepend (key events dropped)

**Keymap pacing / resolve** (`keymap.rs`):

| Test | Contract |
|------|----------|
| `pending_creation_defers_next_physical_input_by_one_delay` | Fresh pending sets delay; next physical stays in delay line |
| `physical_inputs_while_delay_active_stay_in_delay_line` | While delay > 0, physical inputs do not enter the session log |
| `first_tick_while_pending_only_clears_delay` | First tick clears delay without dequeuing |
| `tick_when_delay_zero_moves_one_input_into_session_log` | One delay-line input advances into the session log per processing tick |
| `resolve_by_timeout_leaves_delay_line_inputs_queued` | Never-logged delay-line inputs are not part of resolve replay |
| `post_resolve_ticks_drain_delay_line_as_normal_press` | Post-resolve pacing drains the delay line as a normal press |
| `resolve_leaves_never_logged_delay_line_inputs_queued` | Queue tail that never entered the session log survives resolve |
| `scheduled_input_during_pending_records_while_physical_stays_delayed` | Scheduled input records immediately; concurrent physical stays delayed |
| `resolve_via_tick_leaves_delay_zero_and_queues_replay` | Resolve-in-tick ends delay at 0; replay may remain queued |
| `resolve_inside_handle_input_sets_delay_for_next_physical` | Resolve-in-`handle_input` leaves DELAY set (contrast with tick path) |
| `post_resolve_inside_handle_input_defers_next_physical` | After that path, the next physical input is deferred |

Also keeps the existing #578 regression tests (`scheduled_input_during_pending_*`, `physical_input_during_pending_*`).

## Test plan

- [x] `cargo test -p smart-keymap --lib` (96 passed)

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -762,10 +762,38 @@ impl<
         S: key::System<R, Ref = R, Context = Ctx, Event = Ev, PendingKeyState = PKS, KeyState = KS>,
     > Keymap<I, R, Ctx, Ev, PKS, KS, S>
 {
+    pub(crate) fn test_is_pending(&self) -> bool {
+        self.pending_state.is_some()
+    }
+
     pub(crate) fn test_pending_queued_events_len(&self) -> Option<usize> {
         self.pending_state
             .as_ref()
             .map(|pending_state| pending_state.queued_events.len())
+    }
+
+    /// Session-log input events (only `Event::Input` variants),
+    ///  in log order.
+    pub(crate) fn test_pending_session_log_inputs(
+        &self,
+    ) -> Option<heapless::Vec<input::Event, 16>> {
+        self.pending_state.as_ref().map(|pending_state| {
+            let mut inputs = heapless::Vec::new();
+            for ev in pending_state.queued_events.iter() {
+                if let key::Event::Input(ie) = ev {
+                    let _ = inputs.push(*ie);
+                }
+            }
+            inputs
+        })
+    }
+
+    pub(crate) fn test_input_queue_len(&self) -> usize {
+        self.input_queue.len()
+    }
+
+    pub(crate) fn test_input_queue_delay(&self) -> u8 {
+        self.input_queue.delay_counter()
     }
 
     pub(crate) fn test_handle_scheduled_key_event(&mut self, ev: key::Event<Ev>) {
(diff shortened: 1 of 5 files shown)


src/keymap/pending.rs
```diff
@@ -123,3 +123,129 @@ pub(crate) fn dispatch_replayed_events<Ev: Copy + Debug, const N: usize, const Q
...
+    fn input_ev(ie: input::Event) -> key::Event<()> {
+        key::Event::Input(ie)
+    }
+
+    fn key_ev(keymap_index: u16) -> key::Event<()> {
+        key::Event::Key {
+            keymap_index,
+            key_event: (),
+        }
+    }
```
rgoulter: These look like good helpers which would be worth using in other tests.

src/keymap.rs
```diff
@@ -762,12 +762,40 @@ impl<
...
+        self.pending_state.as_ref().map(|pending_state| {
+            let mut inputs = heapless::Vec::new();
+            for ev in pending_state.queued_events.iter() {
+                if let key::Event::Input(ie) = ev {
+                    let _ = inputs.push(*ie);
+                }
+            }
+            inputs
+        })
+    }
```
rgoulter, line 789: Guard these with `cfg(test)`, I'd think?

rgoulter: They already are — the whole `impl Keymap { … test_* … }` block above is under `#[cfg(test)]` (line ~754). GitHub’s review hunk starts mid-impl so the attribute isn’t visible in the diff context. Happy to add per-method `#[cfg(test)]` too if you want belt-and-braces, but it would be redundant with the impl-level guard.

src/keymap/pending.rs
rgoulter: Done: promoted to `input::Event::press` / `release` constructors and switched the pending, input-event-queue, and `pending_resolution_events` unit tests over to them (also using `key::Event::key_event` / `From<input::Event>` instead of local wrappers).

PR #598, opened 2026-07-10
rust: add optional sticky layer timeout

rgoulter:
## Summary

Allow sticky layers to expire after an optional idle timeout, instead of staying active indefinitely until another key is used.

- New config: `config.layered.sticky_timeout` (milliseconds)
- Default is **unset / `None`** — previous behaviour (sticky stays until used)
- When set, timeout starts when the sticky layer modifier is **released without interruption** (sticky committed)
- If no key is pressed before the timeout, the sticky layer deactivates
- Pressing a key before the timeout still applies the sticky layer (and cancels the timeout)

Similar to ZMK sticky-layer [`release-after-ms`](https://zmk.dev/docs/keymaps/behaviors/sticky-layer).

## Motivation

Tapping `K.layer_mod.sticky` activates a layer for the next key. Without a timeout, that sticky activation can linger if the user changes their mind and types nothing for a long time — the next key later is still modified by the sticky layer.

### Config surface

```ncl
{
  config.layered.sticky_timeout = 1000,  # e.g. 1 second
  layers = [
    [K.layer_mod.sticky 1, K.A, K.B],
    [K.TTTT, K.X, K.Y],
  ],
}
```

### Rust (`src/key/layered.rs`)

1. **`layered::Config`** with `sticky_timeout: Option<u16>` (default `None`).
2. Wired through **`composite::Config.layered`** and `Context::from_config`.
3. On sticky-mod **press**: still emits `StickyActivated` (layer becomes sticky); any prior timeout generation is invalidated.
4. On sticky-mod **release** while still sticky (not interrupted into hold): emits **`StickyReleased`**.
5. Context handles `StickyReleased` by arming a scheduled **`StickyTimeout(generation_id)`** when `sticky_timeout` is `Some`.
6. On `StickyTimeout`: if the generation is still current, no key has used the sticky layer yet (`pressed_keymap_index` is `None`), and a sticky layer is still active → deactivate it.
7. Generation ids cancel stale timeouts when sticky is used, re-armed, converted to hold, or otherwise deactivated.

### Nickel

- `keymap-ncl-to-json` / codegen accept `config.layered.sticky_timeout`
- Codegen emits `sticky_timeout: Some(N)` when set, else `..Config::new()`
- Snapshot `expected.rs` files updated for the new `layered:` config field in composite config

### Tests / docs

- Integration tests: timeout expires / still applies before timeout / no config = no timeout
- Cucumber: two illustrative scenarios (docs)

## Test plan

- [x] `cargo test --test rust-integration layered`
- [x] `cargo test --test cucumber-keymap -- -i '**/layer_modifier-sticky*.feature'`
- [x] `make test-ncl`

### DIFF
--- a/src/key/layered.rs
+++ b/src/key/layered.rs
@@ -260,18 +260,54 @@ impl<const LAYER_COUNT: usize> core::fmt::Debug for ActiveLayersDebugHelper<'_, 
     }
 }
 
+/// Configuration for layered keys / sticky layers.
+#[derive(Deserialize, Debug, Clone, Copy, PartialEq)]
+pub struct Config {
+    /// Timeout (ms) after which an unused sticky layer deactivates.
+    ///
+    /// When [None], sticky layers stay active until another key is used.
+    ///
+    /// The timeout starts when a sticky layer modifier is *released*
+    /// without interruption (sticky committed). If another key is pressed
+    /// while sticky is active, the timeout is cancelled / ignored.
+    #[serde(default)]
+    pub sticky_timeout: Option<u16>,
+}
+
+/// Default layered config (no sticky timeout).
+pub const DEFAULT_CONFIG: Config = Config {
+    sticky_timeout: None,
+};
+
+impl Config {
+    /// Constructs a new default [Config].
+    pub const fn new() -> Self {
+        DEFAULT_CONFIG
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// [crate::key::Context] for [LayeredKey] that tracks active layers.
 #[derive(Clone, Copy)]
 pub struct Context<const LAYER_COUNT: usize> {
+    config: Config,
     default_layer: Option<LayerIndex>,
     active_layers: [Activity; LAYER_COUNT],
     // Keymap index which was pressed while a layer was sticky.
     pressed_keymap_index: Option<u16>,
+    // Invalidates pending sticky-timeout events when advanced.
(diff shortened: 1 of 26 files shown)


src/key/layered.rs
```diff
@@ -260,18 +260,55 @@ impl<const LAYER_COUNT: usize> core::fmt::Debug for ActiveLayersDebugHelper<'_,
     }
 }
 
+/// Configuration for layered keys / sticky layers.
+#[derive(Deserialize, Debug, Clone, Copy, PartialEq)]
+pub struct Config {
+    /// Timeout (ms) after which an unused sticky layer deactivates.
+    ///
+    /// When [None], sticky layers stay active until another key is used
+    /// (previous behaviour).
```
rgoulter: Ugh. LLM-ism to say "previous behaviour" (which is fine in prompts), but not useful for *documentation*.

src/key/layered.rs
```diff
@@ -315,24 +370,59 @@ impl<const LAYER_COUNT: usize> Context<LAYER_COUNT> {
...
+                // Previous sticky wait is superseded while this sticky key is held.
+                self.invalidate_sticky_timeouts();
+                key::KeyEvents::no_events()
+            }
+            LayerEvent::StickyReleased => {
+                // Sticky key released without interruption: layer stays sticky.
+                // Arm optional timeout for "no next key pressed".
+                if self.sticky_layer().is_none() {
+                    return key::KeyEvents::no_events();
+                }
```
rgoulter: `if ... { return ... }` goes against this codebase's conventions.

PR #607, opened 2026-07-18
keymap: re-feed nested pending session inputs in FIFO order

rgoulter:
## Summary

- Nested pending re-queue used `Vec::pop`, which reversed session-log inputs (LIFO).
- Re-queue in chronological (FIFO) order so the new pending key re-observes prior inputs as they occurred.
- Unit tests pin FIFO; integration suite was already insensitive to LIFO vs FIFO.

Small independent fix; split out so it can land without the larger pending-local ingest pacing change.

## Test plan

- [x] `cargo test -p smart-keymap --lib`
- [x] `cargo test -p smart-keymap --test rust-integration`
- [ ] CI green

### DIFF
--- a/src/keymap.rs
+++ b/src/keymap.rs
@@ -454,7 +454,8 @@ impl<
                     key::PressedKeyResult::Pending(pks) => {
                         *pending_key_state = pks;
 
-                        // Pending key transitioned to another pending state: replay session log.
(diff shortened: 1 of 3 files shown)


rgoulter:
Not sure why the code went with LIFO? Seems historically that's what the code I wrote did, but it got formalised as LIFO.

(Practically, I think there's only 1 event at a time or so?)

PR #616, opened 2026-07-19
test: check in trybuild stderr for keymap! nickel fail case

rgoulter:
Baseline for smart-keymap-macros compile_fail UI test so trybuild asserts the Nickel contract error instead of writing wip/ and failing.


rgoulter:
(Apparently this wasn't checked in before, which means it's not run automatically)

PR #623, opened 2026-07-20
Fix nested named layer resolution

rgoulter:


ncl/passes/named-layers.ncl
```diff
@@ -282,13 +282,34 @@
...
+        else if keymap_ncl.chorded_aux.is_key k then
+          (keymap_ncl.chorded_aux.map_accum child_f null k).k
+        else if keymap_ncl.layered.is_key k then
+          (keymap_ncl.layered.map_accum child_f null k).k
+        else if keymap_ncl.tap_hold.is_key k then
+          (keymap_ncl.tap_hold.map_accum child_f null k).k
+        else if keymap_ncl.tap_dance.is_key k then
+          (keymap_ncl.tap_dance.map_accum child_f null k).k
+        else
+          k
```
rgoulter: This smells wrong;
why can't map_accum map the function over its children?

PR #672, opened 2026-08-05
Feat/layer mod tap toggle

rgoulter:
### Summary
Adds a tap-toggle layer modifier: hold behaves like a momentary layer switch, and a single uninterrupted tap toggles the layer on or off. Hold-versus-tap is decided by interruption (another key pressed while the modifier is held), the same pattern as sticky layer mods. On press the layer activates immediately so keys resolve on the target layer right away; release restores or inverts activity based on the pre-press state and whether the press was a hold or a tap.

### Changes
- Core: new tap-toggle modifier variant and release semantics, with unit tests
- Nickel: authoring helper, validation, JSON/codegen, named-layer resolution, and layered-key simulation for input index lookup
- Tests: Rust integration coverage for hold and tap paths (including hold while already toggled on)
- Cucumber feature scenarios documenting the behaviour (QMK TT comparison lives only there)

### Test plan
- [ ] Layered core unit tests
- [ ] Rust integration tests for tap-toggle
- [ ] Cucumber scenarios for the tap-toggle feature
- [ ] Nickel checks for layered/layer-modifier codegen
- [ ] `cargo doc -p smart-keymap-core --no-deps` with warnings denied

### DIFF
--- a/features/keymap/key/layer_modifier-tap_toggle.feature
+++ b/features/keymap/key/layer_modifier-tap_toggle.feature
@@ -1,0 +1,88 @@
+Feature: Layer Modifier: Tap Toggle
+
+  The `K.layer_mod.tap_toggle` key combines momentary and toggle layer
+  access (QMK `TT` with `TAPPING_TOGGLE = 1`):
+
+  - holding the key (interrupted by another key) activates the layer while
+    held, like `K.layer_mod.hold`;
+  - tapping the key toggles whether the layer is active, like
+    `K.layer_mod.toggle`.
+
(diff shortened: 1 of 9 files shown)


rgoulter:
I checked; the tests run on my M4 macbook air. Trying increased timeout.

rgoulter:
Increasing timeout didn't help? Hmm. Will have to come back to this later.

rgoulter:
Tests are very slow, huh; 11m -> 17m is terrible.

rgoulter:
Thinking about it some more, this can be covered with tap-hold. Especially with 'eager decision' logic (TBI).

PR #681, opened 2026-08-06
tap_hold: add retro_tap config

rgoulter:
## Summary

When `config.tap_hold.retro_tap` is true, timeout alone never resolves as hold; hold activates only via interrupt, and alone-release always yields tap (ZMK retro-tap).

Rebased onto current master after tap-hold profiles and `hold_trigger_positions` landed. The knob is now a **profile** field (`Profile.retro_tap`), with flat authoring still available as `config.tap_hold.retro_tap` (lowered into `default_profile`).

- `Profile.retro_tap` + hold-resolution timeout behaviour
- Nickel authoring / lowering / codegen wiring
- Unit tests, Rust integration tests, cucumber feature

## Prior stack (landed)

1. ~~hold_trigger_positions~~ — merged via #691
2. **This PR** — retro_tap
3. quick_tap_ms — #682 (still open; needs similar rebased brush-up)

## Test plan

- [x] `cargo test -p smart-keymap-core --lib key::tap_hold`
- [x] `cargo test --test rust-integration retro_tap`
- [x] `just ncl::checks` (includes `check_profile_with_retro_tap`)
- [x] `just fmt` (rustfmt + nickel format)
- [ ] CI green

### DIFF
--- a/smart-keymap-core/src/key/tap_hold.rs
+++ b/smart-keymap-core/src/key/tap_hold.rs
@@ -25,7 +25,7 @@ pub const MAX_EXTRA_PROFILES: usize = 7;
 ///  not a dense mask of every key on the board.
 pub const MAX_HOLD_TRIGGER_POSITIONS: usize = 16;
 
-/// One tap-hold behavior profile (timeout, interrupt flavor, idle gate, hold triggers).
+/// One tap-hold behavior profile (timeout, interrupt flavor, idle gate, hold triggers, retro-tap).
 ///
 /// Profile 0 is [`Config::default_profile`];
 ///  extra profiles live in [`Config::profiles`]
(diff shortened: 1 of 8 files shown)


smart-keymap-core/src/key/tap_hold.rs
```diff
@@ -189,6 +189,14 @@ pub struct Config {
     /// This reduces disruption from unexpected hold resolutions
     ///  when typing quickly.
     pub required_idle_time: Option<u16>,
+
+    /// When true, timeout alone never resolves as hold (ZMK `retro-tap`).
```
rgoulter: Eh. I kinda think the references to other firmware were find when the keymap was simple.
But, when it's so complex, the other references in smart-keymap-core are unnecessary.

smart-keymap-core/src/key/tap_hold.rs
```diff
@@ -189,6 +189,14 @@ pub struct Config {
...
+    ///
+    /// Hold activates only when another key interrupts (per
+    /// [InterruptResponse]); releasing the key alone always yields tap,
+    /// even after the timeout has elapsed.
+    #[serde(default)]
+    pub retro_tap: bool,
```
rgoulter: Hmm. Looking at the implementation, it seems that this `retro_tap` would be better added to the interrupt response variants.

Downside is: it duplicates across every variant, sure. But upside is: it really is a parameter about how interrupt response is handled.

rgoulter:
Closing as redundant with existing config.

ZMK-style retro-tap (timeout alone never hold; alone-release always tap; hold only on interrupt) is already expressible as:

```ncl
config.tap_hold = {
  timeout = null,
  interrupt_response = "HoldOnKeyPress",  # or HoldOnKeyTap
}
```

See `features/keymap/key/tap_hold-config-timeout-none.feature` and `tests/rust/tap_hold/no_timeout.rs`. A dedicated `retro_tap` flag would only duplicate that composition (similar to FAK, which has no named retro-tap either).

If we want ZMK name recognition later, prefer documenting the equivalence over a parallel config bit.

PR #688, opened 2026-08-09
make: document firmware packages for their target triples

rgoulter:
## Summary

`make test-doc` / `just check-quick` (and thus `just test`) fail on **macOS** when documenting stm32f4/rp2040 packages for the host triple:

```text
error: invalid Mach-O section specifier
  --> …/stm32f4-staging-…/src/stm32f401/mod.rs
   | #[link_section = ".vector_table.interrupts"]
```

PAC crates use ELF section names. That is fine on Linux CI host docs; Mach-O rejects them.

Not caused by the tap-hold profiles work. The failure surfaces after `test-doc` was expanded to cover firmware packages (`dcc8c543`). A devenv/fenix bump can make this more obvious but the underlying issue is host vs target.

## Fix

- `test-doc-stm32f4`: `--target=thumbv7em-none-eabihf` (+ existing `--no-default-features`)
- `test-doc-rp2040`: `--target=thumbv6m-none-eabi --no-default-features`

## Test plan

- [x] `make test-doc-stm32f4` on Darwin
- [x] `make test-doc-rp2040` on Darwin  
- [x] `make test-doc` (host + firmware) on Darwin
- [ ] CI green

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,11 @@ test-clippy:
 # Match CI cargo-doc steps (host + firmware packages).
 # Host: .github/workflows/rust.yaml
 # Firmware: rust-stm32f4.yaml, rust-thumbv6m-none-eabi.yaml
+#
+# Firmware packages pull in PAC crates that use ELF `#[link_section =
+# ".vector_table.interrupts"]`. Documenting them for the *host* triple works on
+# Linux CI (ELF) but fails on macOS (Mach-O rejects those section names). Pass
+# the firmware `--target` so docs build for the real target on all hosts.
 .PHONY: test-doc
 test-doc: test-doc-host test-doc-firmware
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,6 +69,11 @@ test-clippy:
 # Match CI cargo-doc steps (host + firmware packages).
 # Host: .github/workflows/rust.yaml
 # Firmware: rust-stm32f4.yaml, rust-thumbv6m-none-eabi.yaml
+#
+# Firmware packages pull in PAC crates that use ELF `#[link_section =
+# ".vector_table.interrupts"]`. Documenting them for the *host* triple works on
+# Linux CI (ELF) but fails on macOS (Mach-O rejects those section names). Pass
+# the firmware `--target` so docs build for the real target on all hosts.
```
rgoulter: This comment is unnecessary; or at least, this is the wrong part of the code to explain it.
(And prefer semantic linebreaks).

PR #704, opened 2026-08-16
keymap: nested pending timeouts from the physical press

rgoulter:
## Summary

Decision timeouts (tap-hold, tap-dance) are counted from the **physical press**, not from when `new_pressed_key` runs.

A chorded tap-hold used to wait for the chord timeout *plus* the inner timeout (default 200+200 = 400ms). After a pending key is replaced (e.g. chorded passthrough), the keymap now folds already-elapsed time into the inner `After` delay. Remaining 0 is `Immediate`, so an expired inner timeout resolves in the same turn.

Also cancels the previous pending key's delayed events so a leftover chord timeout does not steal a tick.

A *resolved chord* whose binding is itself tap-hold is unchanged: that hold is a fresh `new_pressed_key` on the completing (activator) index.

## Test plan

- [x] `just check-quick`
- [x] `cargo test -p smart-keymap-core --lib`
- [x] `cargo test --test rust-integration` for `chorded`, `tap_hold`, `tap_dance`

### DIFF
--- a/smart-keymap-core/src/key.rs
+++ b/smart-keymap-core/src/key.rs
@@ -114,6 +114,21 @@ impl<E: Copy + Debug> KeyEvents<E> {
     }
 }
 
+impl<E, const M: usize> KeyEvents<E, M> {
+    /// Subtract `elapsed` tick/ms units from each delayed event.
+    ///
+    /// [`Schedule::Immediate`] is unchanged.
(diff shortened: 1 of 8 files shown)


rgoulter:
I think this only really affects tap-hold and tap-dance underneath chorded keys... but it has been a point of confusion since those were added.

This is a CHANGE in behaviour.. but becomes more intuitive.

PR #707, opened 2026-08-21
keymap: keep idle and time at physical press for nested pending

rgoulter:
Fixes chorded passthrough inflating tap-hold required_idle/quick_tap.

- Snapshot KeymapContext at the physical press that creates pending (idle, time, recent presses) in Keymap::pending_press_context
- Reuse that snapshot for nested NewPressedKey so required_idle_time and quick_tap_ms are checked against the press, not the outer chord timeout (timeouts already backdated via pke.backdate)
- Motivating test: tests/rust/chorded/nested_idle.rs demonstrates insufficient 50ms remains tap after 200ms chord timeout, sufficient 150ms still holds

Fixes timing issue where chorded timeout was effectively added to tap-hold idle check.

### DIFF
--- a/smart-keymap-core/src/keymap.rs
+++ b/smart-keymap-core/src/keymap.rs
@@ -335,12 +335,6 @@ pub struct Keymap<I: Index<usize, Output = R>, R, Ctx, Ev: Debug, PKS, KS, S> {
     recent_press_count: u8,
     hid_reporter: HIDKeyboardReporter,
     pending_state: Option<pending::PendingState<R, Ev, PKS>>,
-    /// Snapshot of `KeymapContext` at the physical press that created
-    ///  `pending_state`.
-    /// Used for nested replacements (e.g. chorded passthrough to
-    ///  tap-hold) so `required_idle_time` and `quick_tap_ms` are
-    ///  checked against the press, not the outer timeout.
-    pending_press_context: Option<KeymapContext>,
     input_queue: InputEventQueue<{ MAX_QUEUED_INPUT_EVENTS }>,
     callbacks: heapless::LinearMap<KeymapCallback, CallbackFunction, 2>,
 }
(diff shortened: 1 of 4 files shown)


smart-keymap-core/src/keymap.rs
```diff
@@ -335,6 +335,12 @@ pub struct Keymap<I: Index<usize, Output = R>, R, Ctx, Ev: Debug, PKS, KS, S> {
     recent_press_count: u8,
     hid_reporter: HIDKeyboardReporter,
     pending_state: Option<pending::PendingState<R, Ev, PKS>>,
+    /// Snapshot of `KeymapContext` at the physical press that created
+    ///  `pending_state`.
+    /// Used for nested replacements (e.g. chorded passthrough to
+    ///  tap-hold) so `required_idle_time` and `quick_tap_ms` are
+    ///  checked against the press, not the outer timeout.
+    pending_press_context: Option<KeymapContext>,
```
rgoulter, line 343: This ... apparently works.

But it surely seems like way too much to store; especially without other requirements for it.

My feeling is that storing something like "timings at pending" is appropriate, but the whole keymap context? e.g. also includes recent presses.

smart-keymap-core/src/keymap.rs
```diff
@@ -599,15 +608,28 @@ impl<
...
+                            //  idle to the press time.
+                            keymap_context_without_current_press(
+                                self.recent_presses,
+                                self.recent_press_count,
+                                self.idle_time,
+                                self.event_scheduler.schedule_counter,
+                                pressed_modifiers,
+                                keymap_index,
+                            )
+                        };
```
rgoulter, line 632: Interesting.

I wonder if this suggests it's worth looking into or rearranging the keymap context handling and how it relates to pending events.

PR #711, opened 2026-08-29
Document HID modifier codes (0xE0..0xE7) with hex and constants

rgoulter:
Replaces decimal 224..231 magic numbers with self-documenting forms.

Nickel (45a75aa3): key-extensions and tap_hold codegen now use hid.LeftCtrl etc via record lookup with 0xE0..0xE7; checks use 0xE0 LeftCtrl hex where const not available.

Rust (166c9576): adds HID_LEFT_CTRL..HID_RIGHT_GUI constants and updates from_key_code, is_modifier_key_code, as_key_codes, as_byte and tests; no redundant comments after HID const.

### DIFF
--- a/ncl/smart_keys/keyboard/key-extensions.ncl
+++ b/ncl/smart_keys/keyboard/key-extensions.ncl
@@ -1,28 +1,25 @@
 {
   key_extensions = {
     keyboard =
-      (import "hid-usage-keyboard.ncl")
+      let hid = import "hid-usage-keyboard.ncl" in
+      let modifier_by_key_code = {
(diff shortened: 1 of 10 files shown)


ncl/smart_keys/keyboard/key-extensions.ncl
```diff
@@ -1,28 +1,25 @@
...
-                  225 => { left_shift = true },
-                  226 => { left_alt = true },
-                  227 => { left_gui = true },
-                  228 => { right_ctrl = true },
-                  229 => { right_shift = true },
-                  230 => { right_alt = true },
-                  231 => { right_gui = true },
-                },
-            },
-          kc => { key_code = kc },
```
rgoulter, line 24: This is a bad change. :|

PR #721, opened 2026-09-17
Firmware/ch32x just ux improvements

rgoulter:
The `firmware/ch32x` UX is a bit confusing when dealing with multiple boards.

This updates the `justfile` so it generates a `.hex` with the specified board and keymap in its name. Also tries to update the `flash` with a picker if multiple are available.


firmware/ch32x035-usb-device-compositekm-c/justfile
```diff
@@ -8,16 +77,122 @@ _libsmartkeymap:
...
+      if [ -z "$hex" ]; then
+        hex=$(echo "$hexes" | head -n1)
+      fi
+    fi
+  fi
+  test -n "$hex"
+  echo "flashing $hex"
+  wchisp flash "$hex"
+
+# Alias for explicit flash (kept for compat)
```
rgoulter: "kept for compat" sigh.

firmware/ch32x035-usb-device-compositekm-c/justfile
```diff
@@ -8,16 +77,122 @@ _libsmartkeymap:
...
 
-flash: build await-bootloader
-  wchisp flash build/usb-device-compositekm.hex
+# Flash: with args flashes that board/keymap directly,
+#  without args builds current then picks hex.
+#  One hex -> auto, multiple -> fzf or numbered.
+#  `just flash ch32x-60-improved`
+#  `just flash ch32x-60-improved keymap-66key-extend`
+flash board_arg="" keymap_arg="":
+  #!/usr/bin/env bash
```
rgoulter, line 100: This is a *really* hefty Bash script.. I think it should be in its own `.sh` file.

rgoulter:
I think for followup: it's nice if pickers "remember" history, and may be nice if the CH58x firmware could get the same benefit.

